Issue #1, opened 2019-10-23
MVP for topt login flow


brendanheywood:
* [x] on a new login, after login redirect them to a topt setup page, store the wants url
* [x] render a qr code, they scan it. There is a QR code library inside code

```
require_once($CFG->libdir.'/tcpdf/tcpdf_barcodes_2d.php');
$barcode = new TCPDF2DBarcode($qrcodeurl, 'QRCODE');
$image = $barcode->getBarcodePngData(15, 15);
echo html_writer::img('data:image/png;base64,' . base64_encode($image),''));
```

* [x] there needs to be the ability to have multiple topt per account. Each will have a name
* [x] then they type in the topt to prove that they scanned it and then they get logged in. Log an event for this.
* [x] if they try to go elsewhere then they get redirected back (possibly with a grade period we can add later)

* [x] on subsequent logins they get the topt prompt, and enter it and that is stored in the session
* [x] after either code setup, or code entering, redirect to the wants url

PR #24, opened 2019-11-06
Issue1 implementation

golenkovm:
https://github.com/catalyst/moodle-tool_mfa/issues/1

- Show grace period info page if there is no configured factor for given user.
- Use get_enabled_user_factors() method for simplification.
- Authenticate factors one by one.
- Implement email add factor basic methods.


auth.php
```diff
@@ -55,8 +64,17 @@
 
 if ($form->is_submitted()) {
     if ($data = $form->get_data()) {
-        $_SESSION['USER']->tool_mfa_authenticated = true;
-        redirect(new moodle_url($wantsurl));
+        $property = 'factor_'.$factorname.'_authenticated';
+        $_SESSION['USER']->$property = true;
```
brendanheywood: please just use $SESSION or $USER

classes/local/factor/object_factor_base.php
```diff
@@ -77,12 +77,22 @@ public function add_user_factor($data) {
...
-    public function get_user_factors($user) {
+    public function get_all_user_factors($user) {
+        return array();
+    }
+
+    public function get_enabled_user_factor($userid) {
+        $factors = $this->get_all_user_factors($userid);
+        foreach ($factors as $factor) {
+            if ($factor->disabled == 0) {
+                return $factor;
```
brendanheywood: this doesn't look right, are you filtering this array?

classes/local/factor/object_factor.php
```diff
@@ -36,6 +36,7 @@ public function define_add_factor_form_definition_after_data($mform);
     public function define_login_form($mform);
     public function validation($data);
     public function add_user_factor($data);
-    public function get_user_factors($user);
+    public function get_all_user_factors($user);
```
brendanheywood: Can you please add php docs for all of these as you go so it's clear the intent of each of the methods. eg get_all_user_factors() isn't obvious to me what it's doing (yet)

classes/plugininfo/factor.php
```diff
@@ -82,6 +82,48 @@ public static function get_enabled_factors() {
         return $return;
     }
 
+    /**
+     * Finds enabled factors for user by userid.
+     *
+     * @param int $userid user id.
+     * @return array of enabled factors for given user.
+     */
+    public static function get_enabled_user_factors($userid) {
```
brendanheywood: Is this supposed to return all factors for a user, or all factors for this particular type of factor for a particular user?

classes/local/form/login_form.php
```diff
@@ -33,24 +33,40 @@ class login_form extends \moodleform
...
+        return $mform;
     }
 
     public function validation($data, $files) {
         $errors = parent::validation($data, $files);
 
-        $factors = \tool_mfa\plugininfo\factor::get_enabled_factors();
-        foreach ($factors as $factor) {
+        $factor = \tool_mfa\plugininfo\factor::get_factor($data['factor_name']);
+        if ($factor) {
```
brendanheywood: if $factor is not valid lets make that an $error too

classes/plugininfo/factor.php
```diff
@@ -82,6 +82,48 @@ public static function get_enabled_factors() {
...
+     */
+    public static function get_next_user_factor($userid) {
+        $factors = self::get_enabled_factors();
+
+        foreach ($factors as $factor) {
+            $userfactor = $factor->get_enabled_user_factor($userid);
+            if ($userfactor) {
+                $property = 'factor_'.$userfactor->name.'_authenticated';
+                if (empty($_SESSION['USER']->$property) || true !== $_SESSION['USER']->$property) {
+                    return $userfactor->name;
```
brendanheywood: why not just pass back the factor object instead?

factor/email/classes/factor.php
```diff
@@ -31,4 +32,45 @@
 
 class factor extends object_factor_base {
 
+    public function define_add_factor_form_definition($mform) {
+        global $OUTPUT;
```
brendanheywood: I'm confused about this form, we already know their email. Email should be a zero config factor as the user has already signed up with their email and confirmed it

factor/email/db/install.php
```diff
@@ -0,0 +1,31 @@
...
+ * @package     factor_email
+ * @subpackage  tool_mfa
+ * @author      Mikhail Golenkov <golenkovm@gmail.com>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+defined('MOODLE_INTERNAL') || die;
+
+function xmldb_factor_email_install() {
```
brendanheywood: this file probably should not exist

factor/email/db/install.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<XMLDB PATH="admin/tool/mfa/factor/email/db" VERSION="20191106" COMMENT="XMLDB file for Moodle tool/mfa plugin - email factor"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:noNamespaceSchemaLocation="../../../../lib/xmldb/xmldb.xsd"
+>
+    <TABLES>
+        <TABLE NAME="tool_mfa_factor_email" COMMENT="E-Mail Factor data for MFA plugin">
```
brendanheywood: See above, I don't think we need this table

factor/totp/db/install.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<XMLDB PATH="admin/tool/mfa/factor/totp/db" VERSION="20191104" COMMENT="XMLDB file for Moodle tool/mfa plugin"
+<XMLDB PATH="admin/tool/mfa/factor/totp/db" VERSION="20191104" COMMENT="XMLDB file for Moodle tool/mfa plugin - totp factor"
```
brendanheywood: rename this table from tool_mfa_factor_totp to factor_totp

classes/local/factor/object_factor.php
```diff
@@ -36,6 +36,7 @@ public function define_add_factor_form_definition_after_data($mform);
...
+    public function get_enabled_user_factor($user);
```
brendanheywood: I think this should be passed a $user object, not a $userid

factor/email/classes/factor.php
```diff
@@ -31,4 +32,71 @@
...
+        $disabled = (int)!$this->is_enabled();
+
+        $return = array();
+        $return[1] = new \stdClass();
+        $return[1]->id = $id;
+        $return[1]->name = $name;
+        $return[1]->useremail = $useremail;
+        $return[1]->timemodified = $timemodified;
+        $return[1]->timecreated = $timecreated;
+        $return[1]->disabled = $disabled;
```
brendanheywood: This seems a bit redundant? why not just

```
$return = array(
  'id' => 1,
  'name' => $this->name
...
);
```

Also I would mix in $USER->confirmed as part of the enabled

factor/totp/classes/factor.php
```diff
@@ -46,37 +46,19 @@
...
-                }
-            }
-        }
-        return array();
-    }
-
     public function draw_qrcode($secretcode) {
         global $USER;
-        $code = 'otpauth://totp/'.$USER->username.':'.$USER->email.'?secret='.$secretcode.'&issuer=Moodle&algorithm=SHA1&&period=30';
+        $code = 'otpauth://totp/'.$USER->username.'_2:'.$USER->email.'?secret='.$secretcode.'&issuer=Moodle&algorithm=SHA1&&period=30';
```
brendanheywood: typo?

factor/totp/classes/factor.php
```diff
@@ -90,21 +72,63 @@ public function define_add_factor_form_definition($mform) {
...
-            if ($userfactor->disabled != 1) {
-                $mform->addElement('text', 'verificationcode', get_string('verificationcode', 'factor_totp'));
-                $mform->addRule('verificationcode', get_string('required'), 'required', null, 'client');
-                $mform->setType("verificationcode", PARAM_ALPHANUM);
-            }
+        //if (!empty($secretfield)) {
+        $secret = $secretfield->getValue();
+        $qrcode = $this->draw_qrcode($secret);
+
+        $mform->addElement('html', $OUTPUT->heading('Scan QR-Code or enter a key to your Google Authenticator:', 5));
```
brendanheywood: lang keys for these two headings

factor/totp/classes/factor.php
```diff
@@ -90,21 +72,63 @@ public function define_add_factor_form_definition($mform) {
...
+        $factors = $this->get_enabled_user_factors();
+
+        foreach ($factors as $factor) {
+            if ($factor) {
+                $secret = $factor->secret;
+                $hotp = TOTP::create($secret);
+                $otp = $hotp->now();
+
+                if ($data['verificationcode'] != $otp) {
+                    return array('verificationcode' => 'Wrong verification code');
```
brendanheywood: won't this be wrong if you have multiple totp setup?

Shouldn't it default to an error, and as soon as you hit a match you remove the error?

factor/totp/classes/factor.php
```diff
@@ -90,21 +72,63 @@ public function define_add_factor_form_definition($mform) {
...
         }
 
         return $mform;
     }
 
+    public function login_form_validation($data) {
+        $factors = $this->get_enabled_user_factors();
+
+        foreach ($factors as $factor) {
+            if ($factor) {
```
brendanheywood: is this if required?

factor/totp/classes/factor.php
```diff
@@ -90,21 +72,63 @@ public function define_add_factor_form_definition($mform) {
...
+                if ($data['verificationcode'] != $otp) {
+                    return array('verificationcode' => 'Wrong verification code');
+                }
+            }
+        }
+        return array();
+    }
+
     public function get_secret_length() {
         $length = get_config('factor_totp', 'secret_length');
```
brendanheywood: I would actually remove this setting and hard code it to something very secure which is known to work everywhere. The less settings the better

lib.php
```diff
@@ -24,7 +24,8 @@
 defined('MOODLE_INTERNAL') || die();
 
 function tool_mfa_after_require_login() {
-    if (empty($_SESSION['USER']->tool_mfa_authenticated) || true !== $_SESSION['USER']->tool_mfa_authenticated) {
+    global $USER;
+    if (empty($USER->tool_mfa_authenticated) || true !== $USER->tool_mfa_authenticated) {
         if ($GLOBALS['ME'] != '/admin/tool/mfa/auth.php') {
```
brendanheywood: just use $ME same as $USER

auth.php
golenkovm: Corrected

classes/local/factor/object_factor_base.php
golenkovm: Corrected

classes/local/factor/object_factor.php
golenkovm: Will add php docs in further PR

classes/plugininfo/factor.php
golenkovm: I agree that it's not a best choice for method name. Amended to make it more clear

classes/local/form/login_form.php
golenkovm: I re-wrote this block and implemented new parameter `grace_mode`, so now there is no need to check `if ($factor)`

classes/plugininfo/factor.php
golenkovm: Corrected

factor/email/classes/factor.php
golenkovm: Improved email factor as discussed, so no there is no `add_factor_form` for email factor

factor/email/db/install.php
golenkovm: Corrected

factor/email/db/install.xml
golenkovm: Corrected

factor/totp/db/install.xml
golenkovm: Corrected

classes/local/factor/object_factor.php
golenkovm: Removed argument at all. Using USER global variable in methods

factor/email/classes/factor.php
golenkovm: We need other fields to build a user preference table. Also we user `disabled` property in `get_enabled_user_factors()` method of factor base class

factor/totp/classes/factor.php
golenkovm: In general, yes. Will discuss it later. Corrected

factor/totp/classes/factor.php
golenkovm: Corrected

factor/totp/classes/factor.php
golenkovm: You're right. Re-done this method, now it works fine

factor/totp/classes/factor.php
golenkovm: Corrected

factor/totp/classes/factor.php
golenkovm: Will be done in further PR

lib.php
golenkovm: Done

factor/email/classes/factor.php
brendanheywood: the first comment was purely a code formatting issue, it is 14 lines of code which could be 6 lines of code and do the same thing and be easier to read.

golenkovm: It needs to be an array of objects, not an array of arrays.
However, it can be simplified. Thanks

brendanheywood: You can cast arrays to objects like this:

```
$return = array((object) [
  'id' => 1,
  'name' => $this->name,
  'useremail' => $useremail,
   ...
]);
```

https://stackoverflow.com/questions/19272011/how-to-convert-an-array-into-an-object-using-stdclass/19272047

Issue #5, opened 2019-10-23
TOTP setup flow form have the ability to setup additional topt and revoke old ones


brendanheywood:
* [x] Make a new account preferences page:

https://github.com/catalyst/moodle-tool_securityquestions/blob/master/lib.php#L34

* [x] have ability to revoke each factor, which just disabled it so we have an audit trail

* [x] Each setup should have details of the ip it was created from, the time, and a name. It should also show the last time it was used.

golenkovm:
Done by merging https://github.com/catalyst/moodle-tool_mfa/pull/47 into master branch.
Closing this

PR #47, opened 2019-11-13
TOTP flow

golenkovm:
Issue #5 

- Add 'Created from IP' field for TOTP Factor ;
- Add 'Last logon' field for TOTP factor;
- Update TOTP 'Last logon' field when logon is successful;
- Inject user menu if MFA enabled and there are enabled factors;
- Implement TOTP Factor revocation;
- Move last logon to base class and rename to last verified;
- Align fields and methods names to make factors active/revoked;
- Completely remove edit option;
- Hide Last modified column from user preferences page;
- Use moodle_url() to cunstruct URLs;


classes/local/form/preferences_form.php
```diff
@@ -56,7 +56,9 @@ public function define_configured_factors($mform) {
             'factor',
             'devicename',
             'created',
+            'createdfromip',
             'modified',
+            'lastlogon',
```
brendanheywood: Do we really need modified and lastlogin? In practice won't they end up always being the same?

brendanheywood: Actually yes, if we revoke a token that we can no longer access, then we modify it but don't update lastlogin

factor/totp/classes/factor.php
```diff
@@ -175,11 +175,25 @@ public function login_form_validation($data) {
...
     }
 
+    /**
+     * When validation code is correct - update last logon field for given factor.
+     *
+     * @param factor
+     * @return void
+     * @throws \dml_exception
+     */
+    public function update_lastlogon($factor) {
```
brendanheywood: I think this will end up being common to all factors, can you please move this back to the base class

brendanheywood: I'm also thinking it would be best to rename it. Its a subtle distinction but technically we are not logging on, we are verifying a factor. You could require 3 factors and pass 2 but fail 1 and so not actually login.

Likewise when you setup a new factor you are also verifying it but not logging in. So can you:

1) move to base class. OR at least put it in the interface and leave it here
2) rename to update_lastverified
3) because we have a common api in get_all_user_factors() I think it's best to move the {factor_totp} table into {tool_mfa}. I'm in two minds about this, either the api isn't right or the table is in the wrong spot but I'm not sure which is better. Even things like iprange will still update the verify date so I think it needs to be shared.

classes/local/factor/object_factor_base.php
```diff
@@ -215,4 +215,35 @@ public function login_form_definition_after_data($mform) {
...
+     * Override in child class if necessary.
+     *
+     * @return bool
+     */
+    public function has_delete() {
+        return false;
+    }
+
+    /**
+     * Marks factor record as disabled.
```
brendanheywood: I'm not keen on the inconsistent language, we revoke, but then call a method delete, which toggles a field called disabled. Can you make it revoke / revoked / can_revoke / revoke_user_factor  everywhere

classes/local/form/preferences_form.php
```diff
@@ -81,34 +85,42 @@ public function define_configured_factors($mform) {
...
+            $userfactors = $factor->get_enabled_user_factors();
 
             foreach ($userfactors as $userfactor) {
                 $url = "action.php?sesskey=" . sesskey();
                 $edit = "<a href=\"action.php?sesskey=".sesskey()
                     ."&amp;action=edit&amp;factor=$factor->name&amp;factorid=$userfactor->id\">$headers->edit</a>";
-                $revoke = "<a href=\"action.php?sesskey=".sesskey()
-                    ."&amp;action=revoke&amp;factor=$factor->name&amp;factorid=$userfactor->id\">$headers->revoke</a>";
+                if ($factor->has_delete()) {
+                    $revokeparams = "action=revoke&amp;factor=$factor->name&amp;factorid=$userfactor->id";
```
brendanheywood: Please don't construct urls like this manually. Use moodle_url() so they are escaped properly.

The same goes for links, please use html_writer:

html_writer::link($logouturl, get_string('logout'));

classes/plugininfo/factor.php
```diff
@@ -138,8 +138,8 @@ public static function get_factor_actions() {
         $actions[] = 'add';
         $actions[] = 'revoke';
         $actions[] = 'enable';
```
brendanheywood: 1) looks like duplicate 'revoke's
2) is 'enable' wrong?

lang/en/tool_mfa.php
```diff
@@ -34,21 +34,21 @@
...
 $string['created'] = 'Created';
-$string['modified'] = 'Modified';
+$string['createdfromip'] = 'Created from IP';
+$string['lastverified'] = 'Last verified';
 $string['addfactor'] = 'Add factor';
-$string['editfactor'] = 'Edit factor';
+$string['revokefactor'] = 'Revoke factor';
 $string['enablefactor'] = 'Enable factor';
-$string['disablefactor'] = 'Disable factor';
+$string['revokefactor'] = 'Revoke factor';
```
brendanheywood: duplicate string?

brendanheywood: as a general rule, its nice to keep the lang string files sorted and then it makes it easier to pick up issues like this.

For that reason it can also be helpful to have string keys which are 'object''actions', eg 'factorrevoke' 'factorenable' and then they are all grouped nicely. I'll make another card that that https://github.com/catalyst/moodle-tool_mfa/issues/51

classes/plugininfo/factor.php
golenkovm: Revoke - for factor instance (created by user). 
Disable/enable for factor type (admin settings)

Issue #43, opened 2019-11-12
Use html writer or output for all link construction


brendanheywood:
https://github.com/catalyst/moodle-tool_mfa/blob/master/classes/local/form/preferences_form.php#L95-L98

golenkovm:
Done by merging https://github.com/catalyst/moodle-tool_mfa/pull/53 into master branch

Issue #40, opened 2019-11-12
For totp show for each device when the code was last verified


brendanheywood:
The last modified date probably has little value in it's own right as most factors don't really have a meaningful 'edit' concept'. So I'd just reuse this field and update the record every time it is validated.

When editing a totp record for the purposes of renaming a device, it probably makes sense to enforce them reconfirming the code for that specific device too, so the last mod field is serving double duty without needing two fields.

golenkovm:
Done by merging https://github.com/catalyst/moodle-tool_mfa/pull/53 into master branch

Issue #39, opened 2019-11-12
We need to have an explicit order for factors


brendanheywood:
* [x] usual up and down arrows that other plugin types have like auth

golenkovm:
Done by merging https://github.com/catalyst/moodle-tool_mfa/pull/53 into master branch

Issue #31, opened 2019-11-11
Enabling and disabling factors should call add_to_config_log


brendanheywood:
Most admin settings should be logged for free, just check for all calls to set_config which don't do through the admin UI

golenkovm:
Done by merging https://github.com/catalyst/moodle-tool_mfa/pull/53 into master branch

PR #53, opened 2019-11-14
Release candidate 20191118

golenkovm:
**In progress**

Issues #11 #43 #40 #39 #16 #31 

- Rename add factor to setup factor;
- Implement has_setup() factor method;
- Implement total passed weight check;
- Implement factor ordering;
- Add an entry to the config log table when call set_config();

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -42,7 +42,6 @@ $PAGE->set_pagelayout('popup');
 $pagetitle = $SITE->shortname.': '.get_string('mfa', 'tool_mfa');
 $PAGE->set_title($pagetitle);
 
-
 $OUTPUT = $PAGE->get_renderer('tool_mfa');
 
 $params = array('wantsurl' => $wantsurl);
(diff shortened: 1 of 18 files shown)


classes/local/factor/object_factor_base.php
```diff
@@ -35,6 +35,20 @@ abstract class object_factor_base implements object_factor {
...
+     *
+     * 1 - high priority. Factors without login form should go with this priority.
+     *
+     * 2 - medium priority.
+     *
+     * 3 - low priority.
+     *
+     * @var int
+     */
+    public $priority;
```
brendanheywood: I may not have been clear in the spec, this isn't a priority set by the factor code, this should be an order set by the admin and configured. This should be similar to explicitly setting an order for the authentication plugins. Plugins should not have any say in when they fire. By default each plugin won't really have an order, when it is enabled then it will get added to the end of the list. You need to add UI to move them up and down, which is similar to the enable / disable code. Some places in core also allow drag and drop but it's overkill for here to don't bother.

golenkovm: I was following this spec: "for simplicity this is declared by the plugin and cannot be changed".
https://github.com/catalyst/moodle-tool_mfa/issues/11
But, that's ok, it won't take much time to implement this way ^^

brendanheywood: Ah yes sorry. This has evolved quite a bit :/

lib.php
```diff
@@ -145,3 +145,27 @@ function tool_mfa_extend_navigation_user_settings($navigation, $user, $userconte
...
+ * @return bool
+ */
+function tool_mfa_user_passed_enough_factors() {
+    global $USER;
+    $totalweight = 0;
+    $factors = \tool_mfa\plugininfo\factor::get_active_user_factor_types();
+
+    foreach ($factors as $factor) {
+        $property = 'factor_'.$factor->name;
+        if ($USER->$property == 'good') {
```
brendanheywood: should this be $SESSION not $USER?

index.php
```diff
@@ -38,6 +38,7 @@
 
 $action = optional_param('action', '', PARAM_ALPHANUMEXT);
 $factor = optional_param('factor', '', PARAM_ALPHANUMEXT);
+$sesskey = optional_param('sesskey', '', PARAM_RAW);
```
brendanheywood: You don't need this, you don't actually use $sesskey anywhere

auth.php
```diff
@@ -42,43 +42,47 @@
...
 $PAGE->set_title($pagetitle);
 
-
 $OUTPUT = $PAGE->get_renderer('tool_mfa');
 
 $params = array('wantsurl' => $wantsurl);
 $currenturl = new moodle_url('/admin/tool/mfa/auth.php', $params);
 
-if (isset($USER->tool_mfa_authenticated) && $USER->tool_mfa_authenticated) {
+if (isset($SESSION->tool_mfa_authenticated) && $USER->tool_mfa_authenticated) {
```
brendanheywood: still $USER in the second part

Issue #37, opened 2019-11-12
Cancel button not working on login and add factor forms.


golenkovm:
After clicking on Cancel button `$form->is_cancelled()` returns `false` and `$form->is_submitted()` returns `true`. 
This makes Cancel button unworkable.

golenkovm:
Fixed

PR #57, opened 2019-11-18
Fix Cancel button on login and setup forms

golenkovm:
This PR convers issue https://github.com/catalyst/moodle-tool_mfa/issues/37

### DIFF
--- a/action.php
+++ b/action.php
@@ -70,22 +70,25 @@ switch ($action) {
         $OUTPUT = $PAGE->get_renderer('tool_mfa');
         $form = new setup_factor_form($currenturl, array('factorname' => $factor));
 
-        if ($form->is_cancelled()) {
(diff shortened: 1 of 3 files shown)


action.php
```diff
@@ -70,22 +70,25 @@
...
-                    $event->trigger();
-
-                    redirect($returnurl);
-                } else {
-                    print_error('error:setupfactor', 'tool_mfa', $returnurl);
+            $form->is_validated();
+
+            if ($form->is_cancelled()) {
+                redirect($returnurl);
+            } else {
```
brendanheywood: a tiny thing, do you need this extra nested else? less nesting the better. redirect() will exit on the line above

golenkovm: You're right. Let me fix it, plz

PR #73, opened 2019-11-21
Revoke factor confirmation

### LINKED ISSUE #61: Make revoking factors a 2 step with a confirm button, too easy to brick yourself

golenkovm:
Fixes #61


classes/manager.php
```diff
@@ -133,5 +133,23 @@ public static function get_total_weight() {
...
 
+    /**
+     * Checks that provided factorid exists and belongs to current user.
+     *
+     * @param string $factorname
+     * @param int $factorid
+     * @return bool
+     * @throws \dml_exception
+     */
+    public static function is_factorid_valid($factorname, $factorid) {
```
brendanheywood: Can you please pull $USER out and do that as a function param instead

This will make it easier to unit test, and also better later for when we do non-user actions, eg an admin is revoking it

action.php
```diff
@@ -100,13 +105,40 @@
...
+        if ($form->is_submitted()) {
+            $form->is_validated();
+
+            if ($form->is_cancelled()) {
+                redirect($returnurl);
+            }
+
+            if ($form->get_data()) {
+                if ($factorobject->revoke_user_factor($factorid)) {
+                    $event = \tool_mfa\event\user_revoked_factor::user_revoked_factor_event($USER, $factorobject->get_display_name());
```
brendanheywood: Wouldn't it be better to emit the event inside revoke_user_factor ?

PR #84, opened 2019-11-25
Added factor check to start of auth flow for an early redirect

### LINKED ISSUE #79: Not exiting early when passed
image

Peterburnett:
Closes #79

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -47,6 +47,11 @@ $OUTPUT = $PAGE->get_renderer('tool_mfa');
 
 $currenturl = new moodle_url('/admin/tool/mfa/auth.php');
 
+// Check if enough factors have been passed from no input factors.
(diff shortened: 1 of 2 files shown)


auth.php
```diff
@@ -47,6 +47,11 @@
 
 $currenturl = new moodle_url('/admin/tool/mfa/auth.php');
 
+// Check if enough factors have been passed from no input factors.
+if (tool_mfa_user_passed_enough_factors()) {
+    $SESSION->tool_mfa_authenticated = true;
```
brendanheywood: This looks like we are duplicating, but only partially, logic which is here:

https://github.com/catalyst/moodle-tool_mfa/blob/master/auth.php#L95-L107

Perhaps refactor both of those into a new manager method?

PR #85, opened 2019-11-26
Fallback factor implementation

Peterburnett:


lang/en/tool_mfa.php
```diff
@@ -43,10 +43,13 @@
 $string['error:revoke'] = 'Can not revoke factor';
 $string['error:notenoughfactors'] = 'Not enough factors to authenticate';
 $string['error:factornotenabled'] = 'MFA Factor \'{$a}\' not enabled';
+$string['error:notenoughsetup'] = 'Not enough factors setup to login';
```
brendanheywood: looks like duplicate of 

$string['error:notenoughfactors'] = 'Not enough factors to authenticate';

classes/plugininfo/factor.php
```diff
@@ -239,4 +239,9 @@ public function load_settings(\part_of_admin_tree $adminroot, $parentnodename, $
 
         $adminroot->add($parentnodename, $settings);
     }
-}
\ No newline at end of file
+
+    public static function get_fallback_factor() {
+        return new \tool_mfa\local\factor\fallback();
```
brendanheywood: does a 1 liner function add that much value?

classes/local/factor/fallback.php
```diff
@@ -0,0 +1,52 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Fallback factor class.
+ *
+ * @package     tool_mfa
+ * @author      Mikhail Golenkov <golenkovm@gmail.com>
```
brendanheywood: :)

PR #86, opened 2019-11-26
Issue #65: Refactor auth flow

### LINKED ISSUE #65: Refactor the login flow form
* [x] pass the factor into the form instead of the factor name
* [x] iron out the kinks in the factor / next factor and redirect logic
* [x] likely move the 'grace' or at least the 'failure' fallback modes into pseudo factors https://github.com/catalyst/moodle-tool_mfa/issues/25

Peterburnett:
Closes #65  and #79.

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -47,52 +47,28 @@ $OUTPUT = $PAGE->get_renderer('tool_mfa');
 
 $currenturl = new moodle_url('/admin/tool/mfa/auth.php');
 
-// OVERALL LOGIC
-// 1) Check state DONE
-//      -- If pass, set pass state -> redirect on
-//      -- If fail, cannot_login
-//      else v
-// 2) get the next factor DONE
-// 2.5) Check if fallback factor -> cannot_login DONE
-// 3) Display form for factor
-// 4) Submit form
-// 5) Submit actions ??
-
-
+// Check for overall state.
 $state = \tool_mfa\manager::check_status();
 if ($state == \tool_mfa\plugininfo\factor::STATE_PASS) {
+    unset($SESSION->wantsurl);
     redirect(new moodle_url($wantsurl));
 } else if ($state == \tool_mfa\plugininfo\factor::STATE_FAIL) {
     \tool_mfa\manager::cannot_login();
 }
 
+// If not pass/fail, check next factor state.
 $factor = \tool_mfa\plugininfo\factor::get_next_user_factor();
-
-if ($factor->get_state == \tool_mfa\plugininfo\factor::STATE_FAIL) {
-    // Catch any instant fail state not caught by manager (fallback)
+if ($factor->get_state() == \tool_mfa\plugininfo\factor::STATE_FAIL) {
+    // Catch any instant fail state not caught by manager (fallback).
     \tool_mfa\manager::cannot_login();
 }
 
-/*$userfactors = \tool_mfa\plugininfo\factor::get_active_user_factor_types();
-if (count($userfactors) > 0) {
-    $nextfactor = \tool_mfa\plugininfo\factor::get_next_user_factor();
-    $gracemode = false;
-    $factorname = $nextfactor ? $nextfactor->name : null;
-} else {
-    $factorname = null;
-    $gracemode = true;
-}*/
-
-
-$form = new login_form($currenturl, array('factor' => $nextfactor));
-/*if (isset($nextfactor)) {
-    $factor = \tool_mfa\plugininfo\factor::get_factor($factorname);
-}*/
-
-// KEEP
+// If not, perform form actions for input factor.
+$form = new login_form($currenturl, array('factor' => $factor));
 if ($form->is_submitted()) {
     $form->is_validated();
 
+    // Set state from user actions.
     if ($form->is_cancelled()) {
         $factor->set_state(\tool_mfa\plugininfo\factor::STATE_NEUTRAL);
     } else {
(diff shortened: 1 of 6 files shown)


classes/manager.php
```diff
@@ -144,5 +144,81 @@ public static function is_factorid_valid($factorname, $factorid, $user) {
...
+     */
+    public static function check_status() {
+        global $SESSION;
+
+        // Check for any instant fail states.
+        $factors = \tool_mfa\plugininfo\factor::get_active_user_factor_types();
+        $fail = false;
+        foreach ($factors as $factor) {
+            if ($factor->get_state() == \tool_mfa\plugininfo\factor::STATE_FAIL) {
+                $fail = true;
```
brendanheywood: why not just return here instantly?

classes/manager.php
```diff
@@ -144,5 +144,81 @@ public static function is_factorid_valid($factorname, $factorid, $user) {
...
+        if ($totalweight >= 100) {
+            return true;
+        }
+
+        return false;
+    }
+
+    public static function set_pass_state() {
+        global $SESSION, $USER;
+        $SESSION->tool_mfa_authenticated = true;
```
brendanheywood: Just wondering if we should test if we are already passed and not emit the event the second time?

classes/manager.php
```diff
@@ -144,5 +144,81 @@ public static function is_factorid_valid($factorname, $factorid, $user) {
...
+
+        // Else return neutral state.
+        return \tool_mfa\plugininfo\factor::STATE_NEUTRAL;
+    }
+
+    /**
+     * Checks whether user has passed enough factors to be allowed in
+     */
+    public static function passed_enough_factors() {
+        $totalweight = self::get_total_weight();
```
brendanheywood: I think we should protect this with a fail check, it seems possible to have 100 points and a fail at the same time. Might be moot but can't hurt

auth.php
```diff
@@ -47,86 +47,44 @@
...
-    $gracemode = true;
-}
-
-$form = new login_form($currenturl, array('factor_name' => $factorname, 'grace_mode' => $gracemode));
-if (isset($nextfactor)) {
-    $factor = \tool_mfa\plugininfo\factor::get_factor($factorname);
+// If not pass/fail, check next factor state.
+$factor = \tool_mfa\plugininfo\factor::get_next_user_factor();
+if ($factor->get_state() == \tool_mfa\plugininfo\factor::STATE_FAIL) {
+    // Catch any instant fail state not caught by manager (fallback).
```
brendanheywood: Is this possible?

auth.php
```diff
@@ -47,86 +47,44 @@
 
 $currenturl = new moodle_url('/admin/tool/mfa/auth.php');
 
-if (isset($SESSION->tool_mfa_authenticated) && $SESSION->tool_mfa_authenticated) {
+// Check for overall state.
+$state = \tool_mfa\manager::check_status();
```
brendanheywood: 'check_status' implies we are doing some checking, but it isn't. This is more of a 'get_status'

I'm thinking it would be clearer to have 2 methods, check_status which also does the redirect / cannot_login, and it internally calls get_status. I'm pretty sure we'll want to call check_status in a couple places

auth.php
```diff
@@ -47,86 +47,44 @@
...
-        if (!empty($SESSION->wantsurl)) {
-            unset($SESSION->wantsurl);
-        }
-        redirect(new moodle_url($wantsurl));
-    }
-
-    tool_mfa_logout();
-    print_error('error:notenoughfactors', 'tool_mfa', new moodle_url('/'));
+    // Move to next factor.
+    redirect($currenturl);
```
brendanheywood: At this point we may have just set the current factor to passed, and so now the whole thing is a pass. In which case it would be better to directly redirect to the wants urls, instead of reload this page which then redirects after check_status() at the top. At the same time we could also have a neutral which means we now fail overall and we should just show the error instead of redirecting and then showing the error.

So how about we just call check_status() here (when it is actually doing a check not a get) and you can pass it a boolean to see if you want it to redirect to auth.php or not.

classes/local/form/login_form.php
```diff
@@ -34,21 +34,12 @@ class login_form extends \moodleform {
      */
     public function definition() {
         $mform = $this->_form;
-        $factorname = $this->_customdata['factor_name'];
-        $gracemode = $this->_customdata['grace_mode'];
+        $factor = $this->_customdata['factor'];
 
-        $mform->addElement('hidden', 'factor_name', $factorname);
+        $mform->addElement('hidden', 'factor_name', $factor->name);
```
brendanheywood: Do we need this at all? This feels like a potential vector

classes/local/form/login_form.php
```diff
@@ -104,10 +69,8 @@ public function define_grace_period_page($mform) {
     public function validation($data, $files) {
         $errors = parent::validation($data, $files);
 
-        if (!$data['grace_mode']) {
-            $factor = \tool_mfa\plugininfo\factor::get_factor($data['factor_name']);
-            $errors += $factor->login_form_validation($data);
-        }
+        $factor = \tool_mfa\plugininfo\factor::get_factor($data['factor_name']);
```
brendanheywood: We don't seem to even need the $factor or the factor_name here? dead code from before?

classes/manager.php
```diff
@@ -94,7 +94,7 @@ public static function display_debug_notification() {
             );
         }
 
-        $finalstate = tool_mfa_user_passed_enough_factors()
+        $finalstate = self::passed_enough_factors()
```
brendanheywood: can't we just call check_state (or get_state rather) ?

classes/manager.php
```diff
@@ -144,5 +144,81 @@ public static function is_factorid_valid($factorname, $factorid, $user) {
...
+    }
+
+    /**
+     * Logout user.
+     *
+     * @return void
+     */
+    public static function mfa_logout() {
+        $authsequence = get_enabled_auth_plugins();
+        foreach ($authsequence as $authname) {
```
brendanheywood: What is this for?

I can see prelogout_hook and prelogout_hook but no logoutpage_hook

more dead code from before?

classes/manager.php
```diff
@@ -144,5 +144,81 @@ public static function is_factorid_valid($factorname, $factorid, $user) {
...
+        }
+        require_logout();
+    }
+
+    /**
+     * Function to check the overall status of a user's authentication.
+     *
+     * @return mixed a STATE variable from plugininfo
+     */
+    public static function check_status() {
```
brendanheywood: see above, I think have check_status(boolean $shouldreload) + get_status()

auth.php
Peterburnett: It is possible, currently only in the case of fallback. The fallback will only be caught here, as it doesnt appear as a factor in check_state, otherwise that would fail everytime. This is for when next_factor returns the fallback factor.

classes/local/form/login_form.php
Peterburnett: Its just to track the factor name, as the validation has to re-get the factor, as it cant be passed in as a parameter

classes/local/form/login_form.php
Peterburnett: The factor is needed, to call the validation injection from the factor class

classes/manager.php
Peterburnett: This hook exists in a couple of auth plugins e.g. mnet, shibboleth etc just found from a grep. I think it might be useful to keep, but this code I didnt write, just moved over from lib.php

auth.php
brendanheywood: Ok so this got me thinking. We set an explicit order to the factors. Should we care that a future factor has failed? If we only check them in the correct order we should never get to it and it shouldn't fail overall

The main use case for a hard fail is an explicit deny of an email or a push notification factor, neither of which you would get until you are up to that factor, so I can't see a down side for the real factors. If you hypothetically had an iprange with a black list, then you'd simply order that factor to be first.

classes/manager.php
brendanheywood: ah yes I missed this as it's called from the logout.php

Issue #50, opened 2019-11-14
Reafactor get_all_user_factors() to return an array of factor instances vs factor 'type's


brendanheywood:
* [x] refactor all the factors to use the same db table
* [x] maybe swap from a db record to an object, but I think the db is actually the bigger win here and paves the way for the reporting in https://github.com/catalyst/moodle-tool_mfa/issues/2

PR #88, opened 2019-11-27
Issue #25: factor_grace

Peterburnett:


factor/grace/classes/factor.php
```diff
@@ -0,0 +1,117 @@
...
+        return [$factor];
+    }
+
+    /**
+     * Grace Factor implementation.
+     * Factor cannot be revoked, no extra filtering required.
+     *
+     * {@inheritDoc}
+     */
+    public function get_active_user_factors() {
```
brendanheywood: can we remove this? the base method looks equivalent

factor/grace/lang/en/factor_grace.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Language strings.
+ *
+ * @package     factor_grace
+ * @author      Peter Burnett <peterburnett@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+$string['pluginname'] = 'Grace Period';
```
brendanheywood: watch camel case

factor/grace/classes/factor.php
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public function get_state() {
+        global $USER;
+        $starttime = get_user_preferences('factor_grace_first_login', null, $USER);
+
+        // If no start time is recorded, status is unknown.
+        if (empty($starttime)) {
+            return \tool_mfa\plugininfo\factor::STATE_UNKNOWN;
+        } else {
+            $duration = get_config('factor_grace', 'graceperiod');
```
brendanheywood: also protect against this not being setup?

classes/manager.php
```diff
@@ -263,6 +263,14 @@ public static function set_pass_state() {
             $event = \tool_mfa\event\user_passed_mfa::user_passed_mfa_event($USER);
             $event->trigger();
         }
+
+        // Fire specific factor pass state actions.
+        $pluginsfunction = get_plugins_with_function('mfa_post_pass_state');
```
brendanheywood: 1) why not just a method in the factor and not a new callback hook?

2) Should this be moved above inside the if so it doesn't fire duplicates?

factor/grace/classes/factor.php
```diff
@@ -0,0 +1,111 @@
...
+     *
+     * {@inheritDoc}
+     */
+    public function get_all_user_factors() {
+
+        $factor = (object) array(
+            'id' => 1,
+            'name' => $this->name,
+            'devicename' => '-',
+            'timecreated' => '-',
```
brendanheywood: It just occurred to me that it would be nicer to turn this into a real record and use it to internally store the date they first logged in instead of the user prefs, along side the other bits like their ip and device at the time

factor/grace/lib.php
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @package     factor_grace
+ * @author      Peter Burnett <peterburnett@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+defined('MOODLE_INTERNAL') || die();
+
+function factor_grace_after_require_login() {
```
brendanheywood: Do we need this at all? Can't all this logic be inside get_state()?

classes/local/factor/object_factor_base.php
```diff
@@ -343,4 +343,12 @@ public function create_event_after_factor_setup($user) {
         $event = \tool_mfa\event\user_setup_factor::user_setup_factor_event($user, $this->get_display_name());
         $event->trigger();
     }
+
+    /**
+     * Function for factor actions in the pass state.
+     * Override in child class if necessary.
+     */
+    public function post_pass_state() {
+        return null;
```
brendanheywood: null != void

factor/grace/classes/factor.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Grace Factor implementation.
+     * This factor is a singleton, return single instance.
+     *
+     * {@inheritDoc}
+     */
+    public function get_all_user_factors() {
+        global $DB, $USER;
+
+        if (!$DB->record_exists('factor_grace', array('userid' => $USER->id))) {
```
brendanheywood: There is nothing bad about this but I I think this method I think could be streamlined and be a little more idiomatic moodle code.

insert_record will return the id of the new record, so they way I'd do this is:

1) get_records with s so we get an array (not record_exists), if it exists then we have the full record and we're done, we can just return the array as is and exit early
2) if it is null then build the new record
3) then insert it and set id the id, now we have the same as above with no need for an extra db call, wrap in array and return
4) I would return the record as is with no translation, ie just same all those default values into the record

I'm still a bit weirded out by each factor instance having their own tables which are all the same, but we'll worry about that in #50 

The more I thin about it the more we definitely want a unified table for all factors and for all the factors, even the no input one, to still create and update real records because we need to report on how much and how recently each factor has been used.

factor/grace/classes/factor.php
```diff
@@ -0,0 +1,158 @@
...
+     * Grace Factor implementation.
+     * Checks the user login time against their first login after MFA activation.
+     *
+     * {@inheritDoc}
+     */
+    public function get_state() {
+        // Check if user already has a record, if not, create one at current time.
+        global $USER, $DB;
+
+        if (!$DB->record_exists('factor_grace', array('userid' => $USER->id))) {
```
brendanheywood: can't we just call get_all_user_factors() here and up the first and only record in that array?

factor/grace/classes/factor.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * Grace Factor implementation.
+     * State cannot be set. Return true.
+     *
+     * {@inheritDoc}
+     */
+    public function set_state($state) {
+        return true;
```
brendanheywood: Not that we are actually using the return value at all, but typically most api's return true if it was set. Made https://github.com/catalyst/moodle-tool_mfa/issues/95 to handle this

factor/grace/classes/factor.php
```diff
@@ -0,0 +1,158 @@
...
+
+    public function post_pass_state() {
+        global $SESSION, $USER;
+
+        if (isset($SESSION->grace_message_shown) && $SESSION->grace_message_shown) {
+            return;
+        }
+
+        // Ensure grace factor passed before displaying notification.
+        $grace = \tool_mfa\plugininfo\factor::get_factor('grace');
```
brendanheywood: Can't we just use $this here instead of another lookup to get $grace which should be this same object?

factor/grace/classes/factor.php
```diff
@@ -0,0 +1,158 @@
...
+     * {@inheritDoc}
+     */
+    public function set_state($state) {
+        return true;
+    }
+
+    public function post_pass_state() {
+        global $SESSION, $USER;
+
+        if (isset($SESSION->grace_message_shown) && $SESSION->grace_message_shown) {
```
brendanheywood: I'm wondering if this is needed and is a hangover from before the de-duplication in the calling code and subsequent refactor?

factor/grace/classes/factor.php
Peterburnett: forgot I was inside of the class

factor/grace/classes/factor.php
Peterburnett: A little bit of translation is still required to add 'revoked', however this will be addressed in #50

factor/grace/classes/factor.php
```diff
@@ -0,0 +1,147 @@
...
+
+class factor extends object_factor_base {
+
+    /**
+     * Grace Factor implementation.
+     * This factor needs no user setup, return true.
+     *
+     * {@inheritDoc}
+     */
+    public function setup_user_factor($data) {
```
brendanheywood: do we need this at all?

factor/grace/classes/factor.php
```diff
@@ -0,0 +1,147 @@
...
+            return $records;
+        } else {
+            // Null records returned, build new record.
+            $record = array(
+                'userid' => $USER->id,
+                'createdfromip' => $USER->lastip,
+                'timecreated' => time(),
+            );
+            $record['id'] = $DB->insert_record('factor_grace', $record, true);
+            return $record;
```
brendanheywood: should this be an array of a single record?

Nit: I'd also personally drop the redundant else

PR #91, opened 2019-11-29
IP range factor

### LINKED ISSUE #4: Have an 'iprange' factor
eg you can only set a new TOPT on the cat network the first time you set it up.

* [x] have admin setting , there is an ip range 

address_in_subnet(getremoteaddr()

remoteip_in_list()

Peterburnett:
Closes #4.

### DIFF
--- a/factor/iprange/classes/factor.php
+++ b/factor/iprange/classes/factor.php
@@ -34,7 +34,7 @@ class factor extends object_factor_base {
     /**
      * IP Range Factor implementation.
      * This factor needs no user setup, return true.
-     * 
+     *
      * {@inheritDoc}
      */
     public function setup_user_factor($data) {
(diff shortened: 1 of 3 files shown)


factor/iprange/classes/factor.php
```diff
@@ -31,4 +31,73 @@
...
+     * {@inheritDoc}
+     */
+    public function get_all_user_factors() {
+        $factor = (object) array(
+            'id' => 1,
+            'name' => $this->name,
+            'devicename' => '-',
+            'timecreated' => '-',
+            'createdfromip' => '-',
+            'lastverified' => '-',
```
brendanheywood: same as auth, it would be good to make this a real record and store when we last verified this account was on the secure network

brendanheywood: I'll merge regardless, we can circle back to this when we add the reporting

PR #99, opened 2019-12-03
Issue #2: Admin reports

### LINKED ISSUE #2: Any factor: Have an admin report to show those with this factor setup, eg topt
Summary report:
* [x] matrix of auth plugins vs factors, showing a count for who has it setup

Peterburnett:
Fixes #2 and #50.


classes/local/factor/object_factor_base.php
```diff
@@ -349,6 +349,10 @@ public function create_event_after_factor_setup($user) {
      * Override in child class if necessary.
      */
     public function post_pass_state() {
-        return;
+        // Update lastverified for factor.
```
brendanheywood: If you do this here, won't this mean that if you have say TOTP + Email, and you pass TOTP but fail email, this won't get updated?

I'm in two minds about this, strictly speaking that factor was verified even through over all you didn't pass and log in. In practice it's probably moot

factor/auth/classes/factor.php
```diff
@@ -48,28 +48,25 @@ public function setup_user_factor($data) {
...
-     * Factor cannot be revoked, so no extra functionality is needed.
-     *
-     * {@inheritDoc}
-     */
-    public function get_active_user_factors() {
-        return $this->get_all_user_factors();
+        // Null records returned, build new record.
+        $record = array(
+            'userid' => $USER->id,
+            'authtype' => $USER->auth,
```
brendanheywood: Why do you need to store auth type? This can change fairly easily and it would be simpler to not have to track it here

classes/local/factor/object_factor_base.php
Peterburnett: This is still WIP. The idea is input factors will execute a verified when they are verified, as this is a passive mode for no input factors, it defaults to forcing one when passing if the factor passed. Soon I will have a finalised PR and we can run through the workflow fully. Just battling some regressions atm

factor/totp/classes/factor.php
```diff
@@ -240,12 +242,12 @@ public function setup_user_factor($data) {
...
-        $sql = "SELECT id, 'totp' AS name, devicename, secret, timecreated, createdfromip, timemodified, lastverified, revoked
-                  FROM {factor_totp}
-                 WHERE userid = ?
+        $sql = "SELECT id, 'totp' AS name, label, secret, timecreated, createdfromip, timemodified, lastverified, revoked
+                  FROM {tool_mfa}
+                 WHERE userid = ? AND factor = ?
               ORDER BY revoked, timemodified";
 
-        $return = $DB->get_records_sql($sql, array($USER->id));
+        $return = $DB->get_records_sql($sql, array($USER->id, $this->name));
```
brendanheywood: this could probably be a much simpler get_records() vs get_records_sql() now

auth.php
```diff
@@ -53,14 +53,15 @@
     // Set state from user actions.
     if ($form->is_cancelled()) {
         $factor->set_state(\tool_mfa\plugininfo\factor::STATE_NEUTRAL);
+        // Move to next factor.
+        \tool_mfa\manager::check_status(true);
     } else {
         if ($data = $form->get_data()) {
             $factor->set_state(\tool_mfa\plugininfo\factor::STATE_PASS);
+            // Move to next factor.
+            \tool_mfa\manager::check_status(true);
```
brendanheywood: could these 2 calls just be one and moved outside the ifs?

Peterburnett: Yeah This can be structured better

classes/local/factor/object_factor_base.php
```diff
@@ -230,22 +230,31 @@ public function has_revoke() {
...
-    public function revoke_user_factor($factorid) {
+    public function revoke_user_factor($factorid = null) {
         global $DB, $USER;
-
-        if ($DB->set_field('factor_'.$this->name, 'revoked', 1, array('id' => $factorid))) {
-            $event = \tool_mfa\event\user_revoked_factor::user_revoked_factor_event($USER, $this->get_display_name());
-            $event->trigger();
-
-            return true;
+        if (!empty($factorid)) {
```
brendanheywood: Nit: this feels a little duplicated, you combine these in a $params object and then do the set_field and trigger consistently

brendanheywood: set_field always returns true or throws an exception so I thin the outer if is redundant

classes/local/factor/object_factor_base.php
```diff
@@ -256,20 +265,24 @@ public function revoke_user_factor($factorid) {
      * @return bool
      */
     public function has_lastverified() {
-        return false;
+        return true;
```
brendanheywood: I'm wondering whether this public api is now completely redundant? All factors will now always be able to be verified

classes/local/factor/object_factor_base.php
brendanheywood: So is this comment still valid?

auth.php
brendanheywood: ah I saw it was like that but then moved here. So curious why?

classes/manager.php
```diff
@@ -132,9 +132,9 @@ public static function get_total_weight() {
      * @return bool
      * @throws \dml_exception
      */
-    public static function is_factorid_valid($factorname, $factorid, $user) {
+    public static function is_factorid_valid($factorid, $user) {
         global $DB;
-        $recordowner = $DB->get_field('factor_'.$factorname, 'userid', array('id' => $factorid));
+        $recordowner = $DB->get_field('tool_mfa', 'userid', array('id' => $factorid));
 
         if (!empty($recordowner) && $recordowner == $user->id) {
```
brendanheywood: Nit: I've move the `$recordowner == $user->id` into the sql where clause and then simplify this if to just a return

auth.php
Peterburnett: This was causing an issue where the page would attempt to redirect instead of displaying the validation. It still allowed good requests to pass, but caused display issues

factor/totp/classes/factor.php
```diff
@@ -263,21 +259,22 @@ public function has_revoke() {
      *
      * {@inheritDoc}
      */
-    public function has_lastverified() {
+    public function has_setup() {
         return true;
     }
 
+    public function get_device_name($factorid) {
```
brendanheywood: Does this make sense any more? It is now shared and called label, so I think move this to the base class and call it get_label()

...but I'm not convinced we need it at all. We can probably just remove it completely

factor/totp/lang/en/factor_totp.php
```diff
@@ -42,3 +42,4 @@
 $string['setupfactor:key'] = 'Secret key: ';
 $string['verificationcode'] = 'Enter your 6 digit verification code';
 $string['verificationcode_help'] = 'Open your Authenticator app such as Google Authenticator and look for the 6 digit code which matches this site and username';
+$string['factorreport'] = 'TOTP Factor report';
```
brendanheywood: dud? this is duped in mfa

factor_report.php
```diff
@@ -0,0 +1,95 @@
...
+$table->head = $displaynames;
+$table->align = $colclasses;
+
+// Total row.
+$totalrow = array(get_string('total'));
+foreach ($columns as $column) {
+    if ($column == get_string('tableheader', 'tool_mfa')) {
+        continue;
+    }
+    $sql = 'SELECT COUNT(DISTINCT tm.userid)
```
brendanheywood: this is probably fine as is, but I'm curious if this could be rolled up into a single sql for the lot. I suspect this may be a little hog on a large site.

factor_report.php
```diff
@@ -0,0 +1,95 @@
...
+foreach ($authtypes as $authtype) {
+    $row = array();
+    foreach ($columns as $column) {
+        if ($column == get_string('tableheader', 'tool_mfa')) {
+            // Check if we are inside the authtype column.
+            $row[] = \html_writer::tag('b', $authtype);
+            continue;
+        }
+        // Else, write data.
+        $sql = 'SELECT COUNT(DISTINCT tm.userid)
```
brendanheywood: Ah yes, we'd really want to flatten this out.

That said lets ship is and we can double back and treat this as a follow up optimization issue

settings.php
```diff
@@ -28,6 +28,7 @@
 if ($hassiteconfig) {
 
     $ADMIN->add('tools', new admin_category('toolmfafolder', new lang_string('pluginname', 'tool_mfa'), false));
+    $ADMIN->add('reports', new admin_category('toolmfareports', get_string('mfareports', 'tool_mfa')));
```
brendanheywood: Nit: move this down adjacent to the page which is put inside it

lang/en/tool_mfa.php
```diff
@@ -84,3 +84,9 @@
 $string['state:neutral'] = 'Neutral';
 $string['totalweight'] = 'Total weight';
 $string['weight'] = 'Weight';
+$string['timecreated'] = 'Setup time';
+$string['userid'] = 'User ID';
```
brendanheywood: use the code lang whereever possible

lang/en/tool_mfa.php
```diff
@@ -84,3 +84,9 @@
...
+$string['mfareports'] = 'MFA reports';
+$string['factorreport'] = 'All factor report';
+$string['tableheader'] = 'Auth \ Factor';
```
brendanheywood: Nit: This is one of those very rare times when I'd suggest this is actually 2 strings and the \ is fixed in code. Both strings could probably be reused from elsewhere. This top left corner could just as easily be empty too

factor_report.php
```diff
@@ -0,0 +1,95 @@
...
+$table->data[] = $totalrow;
+
+// Auth rows.
+$authtypes = get_enabled_auth_plugins(true);
+foreach ($authtypes as $authtype) {
+    $row = array();
+    foreach ($columns as $column) {
+        if ($column == get_string('tableheader', 'tool_mfa')) {
+            // Check if we are inside the authtype column.
+            $row[] = \html_writer::tag('b', $authtype);
```
brendanheywood: I'd just move this outside the inner for each and ditch the if / continue

factor_report.php
brendanheywood: UX wise I'd move the totals to the bottom

db/install.xml
```diff
@@ -0,0 +1,26 @@
...
+<XMLDB PATH="admin/tool/mfa/db" VERSION="20191203" COMMENT="XMLDB file for Moodle admin/tool/mfa"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="../../../../lib/xmldb/xmldb.xsd"
+>
+  <TABLES>
+    <TABLE NAME="tool_mfa" COMMENT="Table to store factor configurations for users">
+      <FIELDS>
+        <FIELD NAME="id" TYPE="int" LENGTH="10" NOTNULL="true" SEQUENCE="true"/>
+        <FIELD NAME="userid" TYPE="int" LENGTH="10" NOTNULL="true" SEQUENCE="false" COMMENT="User ID"/>
+        <FIELD NAME="authtype" TYPE="char" LENGTH="100" NOTNULL="false" SEQUENCE="false" COMMENT="Users authentication type"/>
```
brendanheywood: This might have got lost in the last set of polish. Do we need auth type at all?

When we report on this we aren't using this anyway

factor/totp/classes/factor.php
Peterburnett: Its current use case is just for action.php, but I think it is reasonable to broaden the scope of it a bit and move it to the base class. It is currently only used in one place, but it will most likely become useful in future factors, most likely email.

PR #100, opened 2019-12-04
Issue #83: Fixed redirect so a user can only access the auth page

### LINKED ISSUE #83: During the auth login flow you must not be able to go to the prefs page
ie only allow the prefs page if you are in the gracemode fallback step

Peterburnett:
Closes #83

### DIFF
--- a/factor/grace/classes/factor.php
+++ b/factor/grace/classes/factor.php
@@ -45,7 +45,7 @@ class factor extends object_factor_base {
         if (!empty($records)) {
             return $records;
         }
-        
+
         // Null records returned, build new record.
         $record = array(
             'userid' => $USER->id,
--- a/lib.php
+++ b/lib.php
@@ -46,7 +46,7 @@ function tool_mfa_after_require_login() {
 
         $clearurl = str_replace($CFG->wwwroot, '', $ME);
 
-        if (strpos($clearurl, '/admin/tool/mfa/') !== 0) {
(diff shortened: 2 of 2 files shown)


lib.php
```diff
@@ -46,7 +46,7 @@ function tool_mfa_after_require_login() {
 
         $clearurl = str_replace($CFG->wwwroot, '', $ME);
 
-        if (strpos($clearurl, '/admin/tool/mfa/') !== 0) {
+        if (strpos($clearurl, '/admin/tool/mfa/auth.php') !== 0) {
```
brendanheywood: Can you please move the wantsurl logic down into here, we should only set it if we do end up actually redirecting

Can you also please refactor this into a 'should_redirect($url)' method and wrap some unit tests around it.

Right now the logic is simple but its going to get more complex. As a starting point hack you account so that user_not_fully_set_up() returns false which will try to redirect you to edit your profile before this callback fires, which will then start an infinite redirect loop here

brendanheywood: Other tests:
* if you are AJAX never redirect
* if you are a webservice all never redirect

factor/grace/classes/factor.php
```diff
@@ -45,7 +45,7 @@ public function get_all_user_factors() {
         if (!empty($records)) {
             return $records;
         }
-        
+
```
brendanheywood: cleanup

Issue #36, opened 2019-11-12
Added a proper url parser and conditions and tests for when we should redirect to mfa setup pages


brendanheywood:
* [x] Make sure you can't subvert this with a url inside a param
* [x] test for when moodle is installed into a subdirectory

brendanheywood:
This has a lot in common with https://github.com/catalyst/moodle-tool_mfa/issues/68 so maybe do them together

Peterburnett:
Closed in #101

PR #101, opened 2019-12-05
Issue #68: Redirection logic for auth

### LINKED ISSUE #68: Triple check logic around tool_mfa_after_require_login
The entire security of this plugin hinges on this. This needs to be bulletproof.

I've scanned require_login for all the weird edge cases that happen before this hook fires:

* [x] Add unit tests with a data provider with urls. Need to make sure you can't craft a url to get into some other page by hacking query params, or file arguments etc

* [x] I would also like special checks for non web things, ajax calls etc. We need to allow enough stuff to work to allow the page to render fully, but you shouldn't be able to call say an arbitrary ajax web service yet.

* [x] test for when a user logs in the first time, but is not fully setup (eg gets redirected to /user/edit.php) BEFORE this hook is called

* [x] test guest access

* [x] test for forced password change logic. this fires BEFORE this hook

* [x] site policy core redirect fires BEFORE

* [x] soft maintenance mode check

* [x] login as checks, see #22

* [x] enrolment plugin redirects

* [x] if we are in forced lockdown, AND you end up on one of these pages where core has redirected, then we cannot redirect to the auth page or we will go in circles. We need to just print an error and die.

Peterburnett:
Closes #68,  #108, #107 and #36.

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -29,6 +29,10 @@ defined('MOODLE_INTERNAL') || die();
 
 class manager {
 
+    const REDIRECT = 1;
+    const NO_REDIRECT = 0;
+    const REDIRECT_EXCEPTION = -1;
+
     /**
      * Displays a debug table with current factor information.
      */
@@ -257,6 +261,10 @@ class manager {
             $event = \tool_mfa\event\user_passed_mfa::user_passed_mfa_event($USER);
             $event->trigger();
 
+            // Unset session vars during mfa auth.
+            unset($SESSION->mfa_redir_referer);
+            unset($SESSION->mfa_redir_count);
+
             // Fire post pass state factor actions.
             $factors = \tool_mfa\plugininfo\factor::get_active_user_factor_types();
             foreach ($factors as $factor) {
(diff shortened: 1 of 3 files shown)


classes/manager.php
```diff
@@ -257,12 +261,101 @@ public static function set_pass_state() {
...
+            return 0;
+        }
+
+        // WS/AJAX check.
+        if (WS_SERVER || AJAX_SCRIPT) {
+            return self::REDIRECT_EXCEPTION;
+        }
+
+        // User/edit.php.
+        $useredit = new \moodle_url('/user/edit.php');
```
brendanheywood: I think this needs to be more like if (user_not_full_setup() then NO_REDIRECT

Test this in profile editing when some ajax calls

classes/manager.php
```diff
@@ -257,12 +261,101 @@ public static function set_pass_state() {
...
+        }
+
+        // Site policy.
+        if (isset($USER->policyagreed) && !$USER->policyagreed &&
+            defined('NO_SITEPOLICY_CHECK') && !NO_SITEPOLICY_CHECK) {
+            return self::NO_REDIRECT;
+        }
+
+        // Soft maintenance mode.
+        if (!empty($CFG->maintenance_enabled)) {
```
brendanheywood: I'd move this one to the top

I think there is also a couple other similar edge cases to test / check during install / upgrade:

if (isset($CFG->adminsetuppending)) {

if (during_initial_install())

if ($CFG->upgraderunning

Some of these might be redundant and overlap

lib.php
```diff
@@ -44,10 +45,12 @@ function tool_mfa_after_require_login() {
             $SESSION->tool_mfa_setwantsurl = true;
```
brendanheywood: I think only do this if you actually redirect, so move into the if below?

lib.php
```diff
@@ -31,8 +31,9 @@
  * @return void
  * @throws \moodle_exception
  */
-function tool_mfa_after_require_login() {
-    global $SESSION, $ME, $CFG;
+function tool_mfa_after_require_login($courseorid = null, $autologinguest = null, $cm = null,
+    $setwantsurltome = null, $preventredirect = null) {
```
brendanheywood: I've never actually used $setwantsurltome but seems like we should honor this flag too?

classes/manager.php
```diff
@@ -257,12 +261,115 @@ public static function set_pass_state() {
...
+
+        // Initial installation.
+        // We get this for free from get_plugins_with_function.
+
+        // Upgrade check.
+        // We get this for free from get_plugins_with_function.
+
+        // Honor prevent_redirect.
+        if ($preventredirect) {
+            return 0;
```
brendanheywood: is this intentional?

PR #109, opened 2019-12-09
Issue #12: Email factor

### LINKED ISSUE #12: Have an 'email' factor
* [x] has_setup() = false
* [x] is_ready() = do they have an email, and validate_email() + over_bounce_threshold()
* [x] login flow, sends as email with a code, you copy the code into the login form.
* [x] the email also ha the ip address and UA and anything else handy that might help identify the malicious user
* [ ] would usually be very low priority, eg last in the list
* [x] the email should also have a 'This wasn't me' link. The one upside of this approach is that even if your email was compromised you might still see the email. If you do then you can click on the link and it will:
  * [x] work even if you yourself are not logged in (ie you were just doing something else and saw it, it's obviously not you, you need to mitigate it ASAP, and your moodle account may be compromised so you can't actually log in anyway)
  * [x] invalidate the code so the malicious person can't use it 
  * [x] even if they had used that code and logged in, when you click that link it will log them out remotely.
  * [x] log an event with details of the ip address, browser UA and anything else
  * [x] perhaps suspend your account until you can look into it more (needs an option for this)

This still isn't perfect as the person checking your email could also delete the email.

To cleanly do this, a factor can also return a state of 'hard fail' in addition to either a 'yes' (ie yes I'm in the ip range), a no (I'm not in the ip range but thats ok). If any factor returns a fail then the login flow is instantly over.

Peterburnett:
Closes #12.

### DIFF
--- a/factor/email/classes/factor.php
+++ b/factor/email/classes/factor.php
@@ -37,16 +37,9 @@ class factor extends object_factor_base {
      * {@inheritDoc}
      */
     public function login_form_definition($mform) {
-        $userfactors = $this->get_active_user_factors();
 
-        if (count($userfactors) > 0) {
-            $mform->addElement('hidden', 'secret');
-            $mform->setType('secret', PARAM_ALPHANUM);
-
-            $mform->addElement('text', 'verificationcode', get_string('verificationcode', 'factor_email'));
-            $mform->setType("verificationcode", PARAM_ALPHANUM);
-        }
-
+        $mform->addElement('text', 'verificationcode', get_string('verificationcode', 'factor_email'));
+        $mform->setType("verificationcode", PARAM_ALPHANUM);
         return $mform;
     }
 
@@ -56,14 +49,7 @@ class factor extends object_factor_base {
      * {@inheritDoc}
      */
     public function login_form_definition_after_data($mform) {
-        $secretfield = $mform->getElement('secret');
-        $secret = $secretfield->getValue();
-
-        if (empty($secret)) {
-            $secret = random_int(100000, 999999);
-            $secretfield->setValue($secret);
-            $this->email_verification_code($secret);
-        }
+        $this->generate_and_email_code();
         return $mform;
     }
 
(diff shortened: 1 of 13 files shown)


factor/email/classes/factor.php
```diff
@@ -121,4 +127,79 @@ public function get_all_user_factors() {
...
+        }
+        return true;
+    }
+
+    private function generate_and_email_code() {
+        global $USER;
+
+        // No code or old code.
+        $timestamp = get_user_preferences('mfa_email_timestamp', null, $USER);
+        $duration = get_config('factor_email', 'duration');
```
brendanheywood: swap this to factor table

factor/email/classes/factor.php
```diff
@@ -56,28 +49,33 @@ public function login_form_definition($mform) {
...
      * Sends and e-mail to user with given verification code.
      *
      */
-    public function email_verification_code($secret) {
+    public static function email_verification_code($secret) {
         global $USER;
         $noreplyuser = \core_user::get_noreply_user();
         $subject = get_string('email:subject', 'factor_email');
         $message = get_string('email:message', 'factor_email', $secret);
+
```
brendanheywood: push thisn into a renderer which calls a template

factor/email/unauth_email.php
```diff
@@ -0,0 +1,87 @@
...
+        // Unset all sessionvars except lockout.
+        unset_user_preference('mfa_email_ip', $user);
+        unset_user_preference('mfa_email_ua', $user);
+        unset_user_preference('mfa_email_timestamp', $user);
+        unset_user_preference('mfa_email_code', $user);
+
+        $message = get_string('email:revokesuccess', 'factor_email', fullname($user));
+    }
+}
+
```
brendanheywood: should set :STATE_FAIL here somewhere

factor/email/classes/factor.php
```diff
@@ -121,4 +112,119 @@ public function get_all_user_factors() {
...
+        if ($DB->record_exists('tool_mfa', array('userid' => $USER->id, 'factor' => 'email', 'revoked' => 1))) {
+            return false;
+        }
+        return true;
+    }
+
+    private function generate_and_email_code() {
+        global $DB, $USER;
+
+        // Get instance that isnt parent email type (label check).
```
brendanheywood: Can you please add a comment here explaining the two types of records the email check

Issue #98, opened 2019-12-03
Turn the mfa fallback error into a nicer error page with clearer help eg support emails


brendanheywood:

Peterburnett:
This is tandem/duplicate of #106. Doing together

PR #113, opened 2019-12-11
Issue #106: New error message page with support links

### LINKED ISSUE #106: Make the UX of running out of factors nicer
* [x] just a nice error massage as a big lang string which can be customized
* [x] inside a notification error box
* [x] still in the secure layout inside auth.php
* [x] Show the support contact details

Peterburnett:
Closes #106 and #98.

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -147,8 +147,12 @@ class manager {
      * @return void
      */
     public static function cannot_login() {
+        global $OUTPUT;
+        echo $OUTPUT->header();
+        echo $OUTPUT->not_enough_factors();
+        echo $OUTPUT->footer();
         self::mfa_logout();
-        print_error('error:notenoughfactors', 'tool_mfa', new \moodle_url('/'));
+        die;
     }
 
     /**
--- a/renderer.php
+++ b/renderer.php
@@ -175,5 +175,39 @@ class tool_mfa_renderer extends plugin_renderer_base {
 
         return $html;
     }
-}
 
+    /**
+     * Generates notification text for display when user cannot login.
+     *
+     * @return string $notification
+     */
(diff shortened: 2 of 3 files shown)


classes/manager.php
```diff
@@ -148,7 +148,7 @@ public static function is_factorid_valid($factorid, $user) {
      */
     public static function cannot_login() {
         self::mfa_logout();
-        print_error('error:notenoughfactors', 'tool_mfa', new \moodle_url('/'));
+        redirect(new \moodle_url('/admin/tool/mfa/auth.php', array('fail' => 1)));
```
brendanheywood: why not jus display the error directly here without the redirect?

renderer.php
```diff
@@ -175,5 +175,39 @@ public function active_factors() {
...
+    /**
+     * Generates notification text for display when user cannot login.
+     *
+     * @return string $notification
+     */
+    public function not_enough_factors() {
+        global $CFG, $SITE;
+
+        $notification = html_writer::tag('h4', get_string('error:notenoughfactors', 'tool_mfa'));
+        $notification .= get_string('error:reauth', 'tool_mfa').'<br>';
```
brendanheywood: probably wrap this in a p instead of <br>

classes/manager.php
```diff
@@ -147,8 +147,12 @@ public static function is_factorid_valid($factorid, $user) {
      * @return void
      */
     public static function cannot_login() {
+        global $OUTPUT;
+        echo $OUTPUT->header();
+        $OUTPUT->not_enough_factors();
```
brendanheywood: OUTPUT methods should return html not echo them, so you can buffer and compose them etc

PR #116, opened 2019-12-11
Issue #102: Bruteforce timer

### LINKED ISSUE #102: Mitigate timed / repeat brute force attacks on mfa
Assume that the persons password has been compromised, now an attacker is looking at the next layer of defence. We need to protect against both lots of mfa attempts within either a single or multiple sessions.

* [x] Every login flow step should cost some small time say 50ms, and that time should ramp up (double?) on each attempt.
* [x] This should be a user pref and apply across all sessions.
* [x] reset on full login
* [x] do this across the board at the mfa level so all factors get it for free
* [x] in addition have retry limits (or this could also be at the mfa level and handled for free, eg max 5 attempts per factor input step. after the first failure say how many failures you have left

Peterburnett:
Closes #102.

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -48,7 +48,16 @@ $factor = \tool_mfa\plugininfo\factor::get_next_user_factor();
 // If ok, perform form actions for input factor.
 $form = new login_form($currenturl, array('factor' => $factor));
 if ($form->is_submitted()) {
-    $form->is_validated();
+    if (!$form->is_validated() && !$form->is_cancelled()) {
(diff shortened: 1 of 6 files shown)


auth.php
```diff
@@ -69,10 +78,19 @@
     }
 }
 
+// Setup lockout notification.
+if (!empty($SESSION->mfa_fail_counter)) {
```
brendanheywood: can you move this chunk down and merge the two ifs?

PR #120, opened 2019-12-13
Issue #105: Partial MFA rollout

### LINKED ISSUE #105: Support partial MFA rollout
Lets say we rollout MFA only for admins, or only for users for a certain auth type.

In this case it doesn't make sense to show the preferences page to everyone else, as they can set it up but they can't actually ever use it.

Peterburnett:
Also allows for MFA to be optional. Closes #105 and #8

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -490,7 +490,13 @@ class manager {
      * @throws \dml_exception
      */
     public static function is_ready() {
-        global $CFG;
+        global $CFG, $USER;
+
+        // Check if user can interact with MFA.
+        $usercontext = \context_user::instance($USER->id);
+        if (!has_capability('tool/mfa:mfaaccess', $usercontext, $USER)) {
+            return false;
+        }
 
         if (!empty($CFG->upgraderunning)) {
             return false;
@@ -502,7 +508,6 @@ class manager {
         }
 
         $enabledfactors = \tool_mfa\plugininfo\factor::get_enabled_factors();
-
         if (count($enabledfactors) == 0) {
             return false;
         }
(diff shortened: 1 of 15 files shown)


classes/manager.php
```diff
@@ -455,4 +455,56 @@ public static function require_auth($courseorid = null, $autologinguest = null,
             }
         }
     }
+
+    /**
+     * If it is possible for a user to setup a factor that could put them into a pass statek, returns true.
```
brendanheywood: typo statek

classes/manager.php
```diff
@@ -455,4 +455,56 @@ public static function require_auth($courseorid = null, $autologinguest = null,
...
+                if ($factor->has_setup()) {
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        // Filter for which have setup (need to be able to revoke already setup factors).
+        $possiblefactors = array_filter($factors, function($element) {
+            return $element->has_setup() ? true : false;
```
brendanheywood: do you need the ternary?

db/access.php
```diff
@@ -0,0 +1,35 @@
...
+ * Definition of MFA sub-plugins (factors).
+ *
+ * @package     tool_mfa
+ * @author      Peter Burnett <peterburnett@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+defined('MOODLE_INTERNAL') || die;
+
+$capabilities = array(
```
brendanheywood: If you touch caps, please bump version

PR #122, opened 2019-12-13
Issue #64: Lib Refactor

### LINKED ISSUE #64: Refactor all global functions out of lib into a manage class eg tool_mfa_user_passed_enough_factors
The only things in lib should be the callback hooks that must be there.

Peterburnett:
Closes #64

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -432,7 +432,7 @@ class manager {
     $setwantsurltome = null, $preventredirect = null) {
         global $SESSION, $ME;
 
-        if (!self::mfa_ready()) {
+        if (!self::is_ready()) {
             // Set session var so if MFA becomes ready, you dont get locked from session.
             $SESSION->tool_mfa_authenticated = true;
             return;
@@ -466,11 +466,15 @@ class manager {
      * @throws dml_exception
      */
     public static function set_factor_config($data, $factor) {
+        $factorconf = get_config($factor);
         foreach ($data as $key => $newvalue) {
-            $oldvalue = get_config($factor, $key);
-            if ($oldvalue != $newvalue) {
+            if (empty($factorconf->$key)) {
+                add_to_config_log($key, null, $newvalue, $factor);
                 set_config($key, $newvalue, $factor);
-                add_to_config_log($key, $oldvalue, $newvalue, $factor);
+
+            } else if ($factorconf->$key != $newvalue) {
+                add_to_config_log($key, $factorconf->$key, $newvalue, $factor);
+                set_config($key, $newvalue, $factor);
             }
         }
         return true;
@@ -485,7 +489,7 @@ class manager {
      * @return bool
      * @throws \dml_exception
      */
-    public static function mfa_ready() {
+    public static function is_ready() {
         global $CFG;
 
         if (!empty($CFG->upgraderunning)) {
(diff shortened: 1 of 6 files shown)


classes/manager.php
```diff
@@ -455,4 +455,105 @@ public static function require_auth($courseorid = null, $autologinguest = null,
...
+     *
+     * @param array $data
+     * @param string $factor
+     *
+     * @return bool true or exception
+     * @throws dml_exception
+     */
+    public static function set_factor_config($data, $factor) {
+        foreach ($data as $key => $newvalue) {
+            $oldvalue = get_config($factor, $key);
```
brendanheywood: Probably not your code, but I'd move this out of the loop. This will force set a cache item and then invalidte it

so grab the whole plugin config in one go at the start, then set each one which will invalidate it without rebuilding it

classes/manager.php
```diff
@@ -455,4 +455,105 @@ public static function require_auth($courseorid = null, $autologinguest = null,
...
+        $order = explode(',', get_config('tool_mfa', 'factor_order'));
+        $key = array_search($factorname, $order);
+
+        switch ($action) {
+            case 'up':
+                if ($key >= 1) {
+                    $fsave = $order[$key];
+                    $order[$key] = $order[$key - 1];
+                    $order[$key - 1] = $fsave;
+                    self::set_factor_config(array('factor_order' => implode(',', $order)), 'tool_mfa');
```
brendanheywood: refactor all these outside the switch?

index.php
```diff
@@ -58,8 +58,8 @@
 switch ($action) {
     case 'disable':
         if (in_array($factor, $enabledfactors)) {
-            tool_mfa_set_factor_config(array('enabled' => 0), 'factor_' . $factor);
-            tool_mfa_change_factor_order($factor, $action);
+            \tool_mfa\manager::set_factor_config(array('enabled' => 0), 'factor_' . $factor);
```
brendanheywood: change_factor_order() can accecpt a disable / enable, seems like we should just user that

Also seems like change_factor_order should be better named as do_factor_action or similar?

classes/manager.php
```diff
@@ -455,4 +455,105 @@ public static function require_auth($courseorid = null, $autologinguest = null,
...
+    /**
+     * Checks if MFA Plugin is enabled and has enabled factor.
+     * If plugin is disabled or there is no enabled factors,
+     * it means there is nothing to do from user side.
+     * Thus, login flow shouldn't be extended with MFA.
+     *
+     * @return bool
+     * @throws \dml_exception
+     */
+    public static function mfa_ready() {
```
brendanheywood: Nit: the mfa prefix is probably redundant here if we were writing this from scratch, I'd call it is_ready

classes/plugininfo/factor.php
```diff
@@ -239,5 +239,21 @@ public function load_settings(\part_of_admin_tree $adminroot, $parentnodename, $
...
-}
 
+    /**
+     * Checks that given factor exists.
+     *
+     * @param string $factorname
+     *
+     * @return bool
+     */
+    public static function factor_exists($factorname) {
```
brendanheywood: Seems like some conceptual cross over with get_factor, wondering if we can merge these, or just call get_factor from here?

PR #128, opened 2019-12-16
Issue #60: Added notification for successful setup of factors

### LINKED ISSUE #60: On successful setup of a factor, just add a green output->notify as you redirect back to user prefs screen

Peterburnett:
Closes #60.

### DIFF
--- a/action.php
+++ b/action.php
@@ -91,7 +91,8 @@ switch ($action) {
             if ($data = $form->get_data()) {
                 if ($factorobject->setup_user_factor($data)) {
                     $factorobject->set_state(\tool_mfa\plugininfo\factor::STATE_PASS);
-                    redirect($returnurl);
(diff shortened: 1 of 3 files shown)


action.php
```diff
@@ -91,7 +91,8 @@
             if ($data = $form->get_data()) {
                 if ($factorobject->setup_user_factor($data)) {
                     $factorobject->set_state(\tool_mfa\plugininfo\factor::STATE_PASS);
-                    redirect($returnurl);
+                    $finalurl = new moodle_url($returnurl, array('setup' => $factorobject->name));
```
brendanheywood: This whole thing can be much simpler, there is api helpers for doing this as part of a redirect, or sticking it into the session and core handles it:

https://docs.moodle.org/dev/Notifications#Adding_notifications

PR #129, opened 2019-12-17
Issue #112: Added lang strings and methods for accessing strings

### LINKED ISSUE #112: 'Summary of good conditions for login' I'd like to tune it to use custom langs strings provided by each factor
* [x] eg instead of

Administrator factor -50 + Authenticator app 100 + Authentication type 100
it would be
'Is not an admin AND auth type of 'saml' AND using a TOTP app

* [x] Update readme to be same https://github.com/catalyst/moodle-tool_mfa#points-and-examples

Peterburnett:
Closes #112

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ This is so you can specify that logging in via say SAML via ADFS which may have 
 
 #### Non-admin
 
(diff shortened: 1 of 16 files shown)


brendanheywood:
good to go

PR #163, opened 2020-02-18
Issue #153: Detect clock skew on user device.

### LINKED ISSUE #153: Detect clock skew on the users device
* [x] If the code is within say +10 time periods then there is definitely an issue with the client clock being set too far in the future, so hard error "You device clock is wrong"
* [x] if the code is within -10 in the past then it could be either a clock issue or they waited too long

Peterburnett:
Closes #153

### DIFF
--- a/factor/totp/classes/factor.php
+++ b/factor/totp/classes/factor.php
@@ -46,6 +46,13 @@ use tool_mfa\local\factor\object_factor_base;
 use OTPHP\TOTP;
 
 class factor extends object_factor_base {
+
+    const TOTP_OLD = 'old';
+    const TOTP_FUTURE = 'future';
+    const TOTP_USED = 'used';
+    const TOTP_VALID = 'valid';
+    const TOTP_INVALID = 'invalid';
+
     /**
      * Generates TOTP URI for given secret key.
      * Uses site name, domain and user name to make GA account look like:
@@ -184,22 +191,70 @@ class factor extends object_factor_base {
     public function login_form_validation($data) {
         $factors = $this->get_active_user_factors();
         $result = array('verificationcode' => get_string('error:wrongverification', 'factor_totp'));
+        $windowconfig = get_config('factor_totp', 'window');
 
         foreach ($factors as $factor) {
             $totp = TOTP::create($factor->secret);
+            // Convert seconds to windows.
+            $window = (int) floor($windowconfig / $totp->getPeriod());
+            $factorresult = $this->validate_code($data['verificationcode'], $window, $totp, $factor);
+            $time = userdate(time(), get_string('systimeformat', 'factor_totp'));
 
-            // First check if this code matches the last verified timestamp.
-            $lastverified = $this->get_lastverified($factor->id);
-            if ($totp->verify($data['verificationcode'], $lastverified, 1)) {
-                return array('verificationcode' => get_string('error:codealreadyused', 'factor_totp'));
-            }
+            switch ($factorresult) {
+                case self::TOTP_USED:
+                    return array('verificationcode' => get_string('error:codealreadyused', 'factor_totp'));
 
-            if ($totp->verify($data['verificationcode'], time(), 1)) {
-                $result = array();
-                $this->update_lastverified($factor->id);
+                case self::TOTP_OLD:
+                    return array('verificationcode' => get_string('error:oldcode', 'factor_totp', $time));
+
+                case self::TOTP_FUTURE:
+                    return array('verificationcode' => get_string('error:futurecode', 'factor_totp', $time));
+
+                case self::TOTP_VALID:
+                    $this->update_lastverified($factor->id);
+                    return array();
+
+                default:
+                    continue(2);
             }
         }
         return $result;
+    }
+
+    /**
+     * Checks the code for reuse, clock skew, and validity.
+     *
+     * @param string $code the code to check.
+     * @param int $window the window to check validity for.
+     * @param TOTP $totp the totp object to check against.
+     * @param stdClass $factor the factor with information required.
+     *
+     * @return const constant with verification state.
+     */
+    public function validate_code($code, $window, $totp, $factor) {
+        // First check if this code matches the last verified timestamp.
+        $lastverified = $this->get_lastverified($factor->id);
+        if ($totp->verify($code, $lastverified, $window)) {
+            return self::TOTP_USED;
+        }
+
+        // The window in which to check for clock skew, 5 increments past valid window.
+        $skewwindow = $window + 5;
+        $pasttimestamp = time() - ($skewwindow * $totp->getPeriod());
+        $futuretimestamp = time() + ($skewwindow * $totp->getPeriod());
+
+        if ($totp->verify($code, time(), $window)) {
+            return self::TOTP_VALID;
+        } else if ($totp->verify($code, $pasttimestamp, $skewwindow)) {
+            // Check for clock skew in the past 10 periods.
+            return self::TOTP_OLD;
+        } else if ($totp->verify($code, $futuretimestamp, $skewwindow)) {
+            // Check for clock skew in the future 10 periods.
+            return self::TOTP_FUTURE;
+        } else {
+            // In all other cases, code is invalid.
+            return self::TOTP_INVALID;
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


factor/totp/lang/en/factor_totp.php
```diff
@@ -38,6 +40,8 @@
 $string['privacy:metadata'] = 'The TOTP Factor plugin does not store any personal data';
 $string['settings:secretlength'] = 'TOTP secret key length';
 $string['settings:secretlength_help'] = 'Generated TOTP secret key string length';
+$string['settings:window'] = 'TOTP verification window';
```
brendanheywood: can you tweak this to make it clear 1 window = 30 seconds

or maybe even change the drop down so it say 30 sec, 1 min, 2 min, 3 mins, 4 min, 5 min

factor/totp/lang/en/factor_totp.php
```diff
@@ -30,6 +30,8 @@
 $string['devicename_help'] = 'This is the device you have an authenticator app installed on. You can setup multiple devices so this label helps track which ones are being used. You should setup each device with their own unique code so they can be revoked separately.';
 $string['error:wrongverification'] = 'Incorrect verification code';
 $string['error:codealreadyused'] = 'This code has already been used to authenticate. Please wait for a new code to be generated, and try again.';
+$string['error:oldcode'] = 'This code appears to be old. It may be out of date, or your authenticator device time may be out of sync. Please verify device time on your authenticator device, and try again.';
```
brendanheywood: I think to make things easier to debug can you echo the current server time + time zone in both of these

factor/totp/classes/factor.php
```diff
@@ -184,19 +184,29 @@ public function login_form_definition($mform) {
     public function login_form_validation($data) {
         $factors = $this->get_active_user_factors();
         $result = array('verificationcode' => get_string('error:wrongverification', 'factor_totp'));
+        $window = get_config('factor_totp', 'window');
```
brendanheywood: I've just realized if you set the window to say 4, then we need to test back 4 windows in time when checking the code reuse in the other pr

Peterburnett: I updated it to use the window in code as well

Peterburnett: Im not convinced that is exactly the correct behaviour. If checking the previous timestamp, with the window, all codes in the window are treated as already used. If using a window of one, we risk missing the actual code that was used.

factor/totp/lang/en/factor_totp.php
```diff
@@ -38,6 +42,8 @@
 $string['privacy:metadata'] = 'The TOTP Factor plugin does not store any personal data';
 $string['settings:secretlength'] = 'TOTP secret key length';
 $string['settings:secretlength_help'] = 'Generated TOTP secret key string length';
+$string['settings:window'] = 'TOTP verification window';
+$string['settings:window_help'] = 'The window of time codes are allowed to be verified, forwards and backwards.';
```
brendanheywood: admin_setting_configduration allows anything, so I think just add some words here that the window will be rounded to the nearest 30 seconds

factor/totp/classes/factor.php
```diff
@@ -184,19 +184,31 @@ public function login_form_definition($mform) {
...
-            if ($totp->verify($data['verificationcode'], $lastverified, 1)) {
+            if ($totp->verify($data['verificationcode'], $lastverified, $window)) {
                 return array('verificationcode' => get_string('error:codealreadyused', 'factor_totp'));
             }
 
-            if ($totp->verify($data['verificationcode'], time(), 1)) {
-                $result = array();
+            $pasttimestamp = time() - (5 * $totp->getPeriod());
+            $futuretimestamp = time() + (5 * $totp->getPeriod());
+            $time = userdate(time()) . ' ' . usertimezone();
```
brendanheywood: timezone should be in the format eg strftime('%F %T %Z'); and throw that into a new lang string like

langconfig.php:46:$string['strftimedaydate'] = '%A, %d %B %Y';

brendanheywood: note we can tune the string so it is just focused on the time and timezone, we don't care about the date really

factor/totp/classes/factor.php
```diff
@@ -184,19 +184,31 @@ public function login_form_definition($mform) {
...
             // First check if this code matches the last verified timestamp.
             $lastverified = $this->get_lastverified($factor->id);
-            if ($totp->verify($data['verificationcode'], $lastverified, 1)) {
+            if ($totp->verify($data['verificationcode'], $lastverified, $window)) {
                 return array('verificationcode' => get_string('error:codealreadyused', 'factor_totp'));
             }
 
-            if ($totp->verify($data['verificationcode'], time(), 1)) {
-                $result = array();
+            $pasttimestamp = time() - (5 * $totp->getPeriod());
```
brendanheywood: you've got 5 as a magic number in 4 places, can you just pull that out into a shared var with a good name

factor/totp/lang/en/factor_totp.php
```diff
@@ -30,6 +30,10 @@
 $string['devicename_help'] = 'This is the device you have an authenticator app installed on. You can setup multiple devices so this label helps track which ones are being used. You should setup each device with their own unique code so they can be revoked separately.';
 $string['error:wrongverification'] = 'Incorrect verification code';
 $string['error:codealreadyused'] = 'This code has already been used to authenticate. Please wait for a new code to be generated, and try again.';
+$string['error:oldcode'] = 'This code appears to be old. It may be out of date, or your authenticator device time may be out of sync. Please verify device time on your authenticator device, and try again.
```
brendanheywood: This code is too old, please verify the time on your authenticator device is correct and try again.
Current system time is {$a}

factor/totp/lang/en/factor_totp.php
```diff
@@ -30,6 +30,10 @@
...
+    Current system time is {$a}.';
+$string['error:futurecode'] = 'This code appears to be a future code. Your authenticator device time may be out of sync. Please verify device time on your authenticator device, and try again.
```
brendanheywood: This code is invalid, please verify the time on your authenticator device is correct and try again.
Current system time is {$a}.

factor/totp/lang/en/factor_totp.php
brendanheywood: How long each code is valid for. You can set this to a higher value as a workaround if your users device clocks are often slightly wrong. Rounded down to nearest 30 seconds which is the time between new generated codes.

factor/totp/lang/en/factor_totp.php
```diff
@@ -30,6 +30,10 @@
 $string['devicename_help'] = 'This is the device you have an authenticator app installed on. You can setup multiple devices so this label helps track which ones are being used. You should setup each device with their own unique code so they can be revoked separately.';
 $string['error:wrongverification'] = 'Incorrect verification code';
 $string['error:codealreadyused'] = 'This code has already been used to authenticate. Please wait for a new code to be generated, and try again.';
+$string['error:oldcode'] = 'This code is too be old. Please verify the time on your authenticator device is correct and try again.
```
brendanheywood: "This code is too old" I think my typo

Peterburnett: Nah its mine, good spot

factor/totp/classes/factor.php
```diff
@@ -184,24 +191,72 @@ public function login_form_definition($mform) {
...
+     */
+    public function validate_code($code, $window, $totp, $factor) {
+        // First check if this code matches the last verified timestamp.
+        $lastverified = $this->get_lastverified($factor->id);
+        if ($totp->verify($code, $lastverified, $window)) {
+            return self::TOTP_USED;
+        }
+
+        // The window in which to check for clock skew.
+        $skewwindow = 5;
```
brendanheywood: thinking out loud, should this instead be:

$skewwindow = $window + X;  ?

factor/totp/lang/en/factor_totp.php
```diff
@@ -38,10 +42,14 @@
...
+$string['settings:window_help'] = 'How long each code is valid for. You can set this to a higher value as a workaround if your users device clocks are often slightly wrong.
+    Rounded down to the nearest 30 seconds, which is the time between new generated codes.';
 $string['setupfactor'] = 'TOTP Factor Setup';
 $string['setupfactor:scan'] = 'Enter secret or scan QR code';
 $string['setupfactor:key'] = 'Secret key: ';
 $string['verificationcode'] = 'Enter your 6 digit verification code';
 $string['verificationcode_help'] = 'Open your Authenticator app such as Google Authenticator and look for the 6 digit code which matches this site and username';
 $string['summarycondition'] = 'using a TOTP app';
 $string['factorsetup'] = 'Setup App';
+$string['systimeformat'] = '%D %T %Z';
```
brendanheywood: I think this is the neatest default:

%l:%M:%S%P %Z

5:17:36pm Australian Eastern Daylight Time

PR #164, opened 2020-02-18
Issue #160: Emit event for failing MFA

### LINKED ISSUE #160: Emit an event when a user fails mfa and gets logged out

Peterburnett:
Closes #160


classes/event/user_failed_mfa.php
```diff
@@ -0,0 +1,94 @@
...
+     * @throws \coding_exception
+     */
+    public static function user_failed_mfa_event($user) {
+
+        $data = array(
+            'relateduserid' => null,
+            'context' => \context_user::instance($user->id),
+            'other' => array (
+                'userid' => $user->id
+            )
```
brendanheywood: can you dump the same stuff that you'd see in the mfa debug mode into this other and then show it in the desc

ie I want to see something like user X failed mfa 'because they didn't know their totp' - can be ugly as sin as long as the data is there

PR #170, opened 2020-03-30
Issue #168: Added check for deterministic states

### LINKED ISSUE #168: Create a configurable way to hide MFA settings from some users
Approach TBD

Peterburnett:
Closes #168

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -617,8 +617,10 @@ class manager {
         self::set_factor_config(array('factor_order' => implode(',', $order)), 'tool_mfa');
     }
 
-    /*
-     * If it is possible for a user to setup a factor that could put them into a pass state, returns true.
+    /**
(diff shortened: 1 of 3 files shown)


classes/manager.php
```diff
@@ -628,6 +628,18 @@ public static function possible_factor_setup() {
         // Get all active factors.
         $factors = \tool_mfa\plugininfo\factor::get_enabled_factors();
 
+        // Check if there are enough factors that can ONLY pass, if so, don't display.
+        $weight = 0;
+        foreach ($factors as $factor) {
+            $states = $factor->possible_states($USER);
+            if (count($states) == 1 && reset($states) == \tool_mfa\plugininfo\factor::STATE_PASS) {
```
brendanheywood: I feel like this loop and the one below should be combined, the intent I think needs more commenting

PR #177, opened 2020-04-22
Issue 176: Improve factor report UI

### LINKED ISSUE #176: MFA report numbers should be more like a funnel
/admin/tool/mfa/factor_report.php

ie the first line might look something like:

. | Non-administrator | IP Range Factor | Authenticator app | Grace period
-- | -- | -- | -- | --
manual | 700 | 700 | 20 | 700

But what we want it to show is really the funnel of which uses hit and passed each step, so probably more like:

. | Users | Logins | Non-administrator | IP Range Factor | Authenticator app | Grace period
-- | -- | -- | -- | -- | -- | --
manual | 1000 | 722 | 700 | 20 | 13 | 7

ie in plain english:

1) 720 manual users have attempted to login
2) 700 of them passed the no-admin factor and so stopped there
3) which means 22 were admins, of those 20 logged in via a secure network at some point
4) of those 13 have setup totp
5) of the remained 7 have been in grace mode

So a few tasks might drop out of this:

* [x] have a button to reset the stats
* [x] show the time at which stats have started being recorded from
* [x] add the starting 2 columns which is the raw number users and logins before we do any factor filtering funnel
* [ ] some stats need to be have the ability to fade away, ie the grace mode should be able to remove people whose grace mode has expired. we want that number to be zero after the roll out
* [x] the report itself and the stats should be recording stats on just the passed / triggered factors not all of them
* [ ] have a drill down report so we can see what users are behind each number

Peterburnett:
Closes #176.


factor_report.php
```diff
@@ -51,21 +51,47 @@
 foreach ($authtypes as $authtype) {
     $row = array();
     $row[] = \html_writer::tag('b', $authtype);
+
+    // Manually handle Total users and logged in users.
+    $totalusercountsql = "SELECT COUNT(DISTINCT id)
+                            FROM {user}
+                           WHERE auth = ?";
```
brendanheywood, line 58: filter deleted? / suspended? probably same below

brendanheywood: also would this be better to move outside the foreach and use a group by auth? same below

factor_report.php
```diff
@@ -30,63 +30,138 @@
...
+    $row[] = $allusersinfo[$authtype]->count ?? '-';
+    $row[] = $mfausersinfo[$authtype]->count ?? '-';
+
+    // This gets all the factors we need information on.
+    $factorsusedssql = "SELECT tm.id as factorid,
+                                tm.userid as id,
+                                tm.factor as factor
+                           FROM {tool_mfa} tm
+                           JOIN {user} u
+                             ON tm.userid = u.id
```
brendanheywood, line 103: super tiny, I think the ON should be up a line

factor_report.php
```diff
@@ -30,63 +30,138 @@
...
+                          WHERE u.auth = ?
+                            AND tm.revoked = 0
+                            AND tm.lastverified >= ?";
+    $records = (array) $DB->get_records_sql($factorsusedssql, [$authtype, $lookback]);
```
brendanheywood, line 107: whats the (array) for?

factor_report.php
```diff
@@ -30,63 +30,138 @@
...
-                        ON tm.userid = u.id
-                    WHERE u.auth = ? AND tm.factor = ?';
-            $count = $DB->count_records_sql($sql, array($authtype, $column));
+            $factoruserrecords = array_filter($records, function($element) use ($column) {
+                return $element->factor === $column;
+            });
+
+            // Now diff the factorusers to remove any users who authed with this factor,
+            // And update the parent array for the next iteration.
+            $newrecords = array_diff_key($records, $factoruserrecords);
```
brendanheywood, line 120: really not sure about this logic

PR #183, opened 2020-04-27
Make connector string able to be localized.

nomisge:
The connector string has trailing white spaces. This is bad practice, because it cannot be translated.
I changed the code, such that the whitespaces are introduced within the code and not the localized string.


Peterburnett:
Hi @nomisge The patch looks great. Are you able to rebase this patch to the latest master to resolve the merge conflict, and squash the commits to 1 commit? After that it should be good to go

Peterburnett:
Merged. Thanks @nomisge for the patch!

PR #192, opened 2020-05-06
Issue #191: Lockout a factor if user exceeds attempts

### LINKED ISSUE #190: Improve UX notification if TOTP lockout is triggered
It should show that this factor can now never be  used to login until a successful reauth

### LINKED ISSUE #191: If an input factor is the factor that triggers and MFA logout, lock that factor from being used again
The factor should become usable again when an OK MFA auth occurs for the user. Related to #190

Peterburnett:
Closes #191. Also closes #190, users will never see failing factor now, and the message makes it clear that the lockout will apply to the factor


classes/local/factor/object_factor_base.php
```diff
@@ -35,14 +35,30 @@ abstract class object_factor_base implements object_factor {
...
      *
      * @param string factor name
      *
      */
     public function __construct($name) {
+        global $DB, $USER;
         $this->name = $name;
+
+        // Setup the lock counter.
+        $sql = "SELECT MAX(lockcounter) FROM {tool_mfa} WHERE userid = ? AND factor = ?";
```
brendanheywood, line 54: should this exclude previously revoked factors?

classes/local/factor/object_factor_base.php
```diff
@@ -433,4 +449,33 @@ public function delete_factor_for_user($user) {
...
+     * Increments the lock counter for a factor.
+     *
+     * @return void
+     */
+    public function increment_lock_counter() {
+        global $DB, $USER;
+
+        $this->lockcounter++;
+        // Update record in DB.
+        $DB->set_field('tool_mfa', 'lockcounter', $this->lockcounter, ['userid' => $USER->id, 'factor' => $this->name]);
```
brendanheywood, line 463: should match revoke exclusions above if we add them

classes/local/factor/object_factor_base.php
Peterburnett: I think the revoked factors should also have this field incremented. We want the lockoutcounter to be in lockstep for the entire class of factor instances, so I think that should include all of them

classes/local/factor/object_factor_base.php
Peterburnett: See above reply

PR #213, opened 2020-06-24
Changed error page to allow for state changes in factors

### LINKED ISSUE #211: Bugs with lang strings on failed mfa check
/admin/tool/mfa/auth.php

You are logged in as Brendan Heywood
[[pluginname]]

image

Peterburnett:
Closes #211

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -37,6 +37,21 @@ $PAGE->set_pagelayout('secure');
 $pagetitle = $SITE->shortname.': '.get_string('mfa', 'tool_mfa');
 $PAGE->set_title($pagetitle);
 
+// The only page action allowed here is a logout if it was requested.
+$logout = optional_param('logout', false, PARAM_BOOL);
+if ($logout) {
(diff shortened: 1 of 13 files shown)


auth.php
```diff
@@ -37,6 +37,14 @@
 $pagetitle = $SITE->shortname.': '.get_string('mfa', 'tool_mfa');
 $PAGE->set_title($pagetitle);
 
+// The only page action allowed here is a logout if it was requested.
+$logout = optional_param('logout', false, PARAM_BOOL);
+if ($logout) {
+    \tool_mfa\manager::mfa_logout();
+    // Now redirect home.
+    redirect(new \moodle_url($CFG->wwwroot));
```
brendanheywood, line 45: Would it be possible to redirect back to the original wants url? so that after the second loop through you still end up in the right place

Peterburnett: Ah thats a nice feature.

PR #217, opened 2020-07-14
Issue #172: Added user capability factor

### LINKED ISSUE #172: Have an exempt from mfa capability

Peterburnett:
Closes #172

### DIFF
--- a/factor/capability/classes/factor.php
+++ b/factor/capability/classes/factor.php
@@ -1,0 +1,111 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * User capability factor class.
+ *
+ * @package     tool_mfa
+ * @author      Peter Burnett <peterburnett@catalyst-au.net>
+ * @copyright   Catalyst IT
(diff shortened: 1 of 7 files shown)


factor/capability/lang/en/factor_capability.php
```diff
@@ -0,0 +1,31 @@
...
+ * @author      Peter Burnett <peterburnett@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+$string['pluginname'] = 'User capability';
+$string['privacy:metadata'] = 'The  user capability factor plugin does not store any personal data';
+$string['settings:adminpasses'] = 'Site admins pass this factor';
+$string['settings:adminpasses_help'] = 'Controls whether site administrators should be able to pass this factor without an explicit role binding.';
+$string['summarycondition'] = 'does NOT have the factor/capability:cannotpassfactor capability';
```
brendanheywood, line 30: can you include the log of the adminpass in this string, possibly split into two strings

factor/capability/lang/en/factor_capability.php
```diff
@@ -0,0 +1,31 @@
...
+ * @package     factor_capability
+ * @author      Peter Burnett <peterburnett@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+$string['pluginname'] = 'User capability';
+$string['privacy:metadata'] = 'The  user capability factor plugin does not store any personal data';
+$string['settings:adminpasses'] = 'Site admins pass this factor';
+$string['settings:adminpasses_help'] = 'Controls whether site administrators should be able to pass this factor without an explicit role binding.';
```
brendanheywood, line 29: I might have got the language here the wrong way around, but I think the help could flesh this out something like:

By default admins pass all capability checks, which means they will fail this factor.

If checked then all site admins and anyone without this capability will pass.

If not checked the only site admins without this capability will pass.

factor/capability/classes/factor.php
```diff
@@ -0,0 +1,111 @@
...
+     *
+     * {@inheritDoc}
+     */
+    public function has_input() {
+        return false;
+    }
+
+    /**
+     * User capability implementation.
+     * Checks a users current IP against allowed and disallowed ranges.
```
brendanheywood, line 73: cut and paste typo

PR #226, opened 2020-08-12
Issue #220: SMS Factor

### LINKED ISSUE #220: Implement an SMS factor using Amazon SNS
There will be many potential sms gateways to use and I'm on the fence around whether it should be 1 factor with subplugins, or different factors. The config should be fairly minimal but will be almost exactly the same, the main difference will be api keys etc for each gateway. So leaning towards a base class and then 1 factor / subplugin for each gateway service

* [x] have config to decide where the phone number comes from. There are already the built in 'Phone' and 'Mobile' phone fields in the core user table
* [ ] or it could be a custom field
* [x] or it could be managed in the factor plugin

I think for MVP we'd want to probably hard code usage of the Mobile phone core field, but with some UX sugar so you can also edit it via the factor setup page.

* [x] api config, should be either api keys or accept the aws roles

* [x] On the factor input step page, there should probably be a half redacted message 'A code has been sent to a phone ending in 04xx xxx 123'. ~~Either that or it is a prompt and the drop down has the Phone + Mobile phone and you choose which to send to~~. If more than 1 phone is registered then send a text to all of them.

Peterburnett:
closes #220


factor/sms/classes/factor.php
```diff
@@ -0,0 +1,241 @@
...
+        // This check must exclude the main singleton record, with the label as the email.
+        // It must only grab the record with the user agent as the label.
+        $sql = 'SELECT *
+                  FROM {tool_mfa}
+                 WHERE userid = ?
+                   AND factor = ?
+               AND label IS NOT NULL';
+        $record = $DB->get_record_sql($sql, array($USER->id, 'sms'));
+
+        if ($enteredcode == $record->secret) {
```
brendanheywood, line 204: I'm not sure we should store the code here. Scenario is a little far fetched but real:

1) phone is stolen by mallory, notifications are shown on the lock screen (bad but not uncommon)
2) alice gets a phishing attack from mallory
3) she logs in to fake site (bad) so now mallory has the password. Up until now its non mfa so mfa should mitigate this
4) mallory gets the sms code and they are in because it works for their session. MFA has not helped

Unlike TOTP we can tie the sms code to the session instead of storing it so now the code is useless to mallory. And we can also go a step further and treat a code use by a different session as suspicious or an explicit fail rather than a neutral

factor/sms/classes/local/smsgateway/gateway_interface.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+namespace factor_sms\local\smsgateway;
+
+defined('MOODLE_INTERNAL') || die();
+
+interface gateway_interface {
+    /**
+     * @param string $messagecontent the content to send in the SMS message.
+     * @param string $target the destination for the message.
```
brendanheywood, line 33: target == phonenumber? maybe rename

factor/sms/settings.php
```diff
@@ -0,0 +1,66 @@
...
+    get_string('settings:aws:key', 'factor_sms'),
+    get_string('settings:aws:key_help', 'factor_sms'),
+    ''));
+
+$settings->add(new admin_setting_configpasswordunmask('factor_sms/api_secret',
+    get_string('settings:aws:secret', 'factor_sms'),
+    get_string('settings:aws:secret_help', 'factor_sms'),
+    ''));
+
+// These are the only regions that  is available in.
```
brendanheywood, line 47: can this be dynamic?

factor/sms/version.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+defined('MOODLE_INTERNAL') || die();
+
+$plugin->version      = 2020081100;      // The current plugin version (Date: YYYYMMDDXX).
+$plugin->requires     = 2017051500.00;   // Support back to 3.3 - Totara 12. Patches required.
+$plugin->component    = 'factor_sms';
+$plugin->release      = 'v0.1';
+$plugin->maturity     = MATURITY_STABLE;
+$plugin->dependencies = array('tool_mfa' => 2019102400, 'local_aws' => 2020061500);
```
brendanheywood, line 34: I dont think this should have a hard dependancy or you wont be able to install mfa even if you don't use sms

factor/sms/classes/factor.php
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+
+    private function sms_verification_code($instanceid) {
+        global $DB;
+
+        // Here we should get the information, then construct the message.
+        $instance = $DB->get_record('tool_mfa', ['id' => $instanceid]);
+        $user = \core_user::get_user($instance->userid);
+        $phonenumber = $user->phone2;
```
brendanheywood, line 180: So I'm quite conflicted about using this profile field vs storing the number in the factor table like totp secret. Pros and cons:

Profile:

pros: it makes sense to the user
cons: zero validation, it can change at any time by the user. It could be the word 'frog' or have a # and extension numbers etc

Factor:

pro: full contol over validation
cons: different UX

So I'm open to all discussion but I'm leaning towards some sort of hybrid:

1) the user can supply the phone (eg on the registration page)
2) the MFA factor never trusts this, it must confirm it before using it the same as totp
3) if the factor does not have a stored number, but it detects that the profile does have one then will send a text to confirm
4) user confirms like totp, at this point the number is cloned into the factor table and we display it as something like 'confirmed number'
5) bonus points we can register event for when the profile fields have changed and compare to the confirmed number, if it is different then we can register a second number and probably leave the previous one in place to be manually removed in the mfa preferences

I think as bonus points the setup_factor_form_definition($mform) should exist and show a form which directly edits the mobile phone profile field, but which also adds some validation to it. The validation doesn't need to be perfect as we will be sending an sms to it very soon after. We should be fairly liberal with accepting ( and ) - etc and removing them. The challenge is we can't do this validation directly on the profile unless we make a sibling custom profile field type which is a whole new plugin which sucks.

This looks promising:

https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch04s03.html

factor/sms/classes/factor.php
```diff
@@ -0,0 +1,241 @@
...
+class factor extends object_factor_base {
+    /**
+     * SMS Factor implementation.
+     *
+     * {@inheritDoc}
+     */
+    public function login_form_definition($mform) {
+
+        $mform->addElement('text', 'verificationcode', get_string('verificationcode', 'factor_sms'),
+            ['autocomplete' => 'one-time-code']);
```
brendanheywood, line 42: also add some more bits:
```
<input
  type="text"
  name="token"
  id="token"
  inputmode="numeric"
  pattern="[0-9]*"
  autocomplete="one-time-code"
/>
```

https://www.twilio.com/blog/html-attributes-two-factor-authentication-autocomplete

factor/sms/classes/factor.php
brendanheywood: except no in this case mallory could replay both the password and also now get the sms text directly so it won't really help. hhmmm maybe overkill

factor/sms/classes/factor.php
```diff
@@ -0,0 +1,241 @@
...
+        return true;
+    }
+
+    /**
+     * SMS Factor implementation.
+     *
+     * {@inheritDoc}
+     */
+    public function has_setup() {
+        return true;
```
brendanheywood, line 116: This looks quite promising:

https://aws.amazon.com/blogs/messaging-and-targeting/improve-sms-message-delivery-with-phone-number-verify/

> With Phone Number Verify, Amazon Pinpoint checks to see whether the phone number provided is in a valid format based on its country code prefix. If the number is formatted correctly, Amazon Pinpoint leaves the number as the customer entered it. If the number isn’t formatted correctly, Amazon Pinpoint compares the phone number to various rules to make it valid. For example, if a customer enters a leading zero after the country code, Phone Number Verify removes the extra character to make the number valid. In this case, if the customer entered +1 0 206 555 0199 (a United States phone number in the 206 area code), Phone Number Verify changes the number to a correctly formatted number without the extra zero (+1 206 555 0199).

brendanheywood: I'll make this a new issue for later

brendanheywood: https://github.com/catalyst/moodle-tool_mfa/issues/227

factor/sms/classes/local/smsgateway/aws_sns.php
```diff
@@ -0,0 +1,61 @@
...
+            'region' => $config->api_region,
+            'credentials' => [
+                'key' => $config->api_key,
+                'secret' => $config->api_secret
+            ],
+            'http' => ['proxy' => \local_aws\local\aws_helper::get_proxy_string()]
+        ]);
+        // These messages need to be transactional.
+        $client->SetSMSAttributes([
+            'attributes' => [
```
brendanheywood, line 48: We should set the sender id

```
          'AWS.SNS.SMS.SenderID' => [
               'DataType' => 'String',
               'StringValue' => 'The Clients Name'
        ]
```

The value should be taken from the support contact CFG

See

https://github.com/aws/aws-sdk-php/issues/1321

https://docs.aws.amazon.com/sns/latest/dg/channels-sms-originating-identities.html

https://github.com/aws/aws-sdk-php/issues/1321

classes/local/factor/object_factor_base.php
```diff
@@ -327,6 +327,15 @@ public function has_setup() {
         return false;
     }
 
+    /**
+     * If has_setup returns true, decides if the setup buttons should be shown on the preferences page.
+     *
+     * @return bool
+     */
+    public function show_setup_buttons() {
+        return false;
```
brendanheywood, line 336: should the default impl just return $this->has_setup() ?

classes/local/form/verification_field.php
```diff
@@ -0,0 +1,70 @@
...
+        // Force attributes.
+        if (empty($attributes)) {
+            $attributes = [];
+        }
+
+        $attributes['autocomplete'] = 'one-time-code';
+        $attributes['autofocus'] = 'autofocus';
+        $attributes['inputmode'] = 'numeric';
+        $attributes['pattern'] = '[0-9]*';
+        $attributes['style'] = '
```
brendanheywood, line 46: can you add this as a class and move the css to .../styles.css which is magically loaded into the main css file, this way a theme can override it

classes/local/form/verification_field.php
```diff
@@ -0,0 +1,70 @@
...
+            font-size:1.25em;
+            letter-spacing:1.05em;
+            font-family:monospace;
+            width:11.5em;
+        ';
+
+        // Load JS for element.
+        $PAGE->requires->js_call_amd('tool_mfa/autosubmit_verification_code', 'init', []);
```
brendanheywood, line 54: in theory this should bind the code directly to this instance, but not a big deal here

classes/local/form/verification_field.php
```diff
@@ -0,0 +1,70 @@
...
+
+        // Force element name to match JS.
+        $elementname = 'verificationcode';
```
brendanheywood, line 57: ah yes, this is kinda the same as above, would be good to not tie this directly

Peterburnett:
> should the default impl just return $this->has_setup() ? 

Yeah it should

factor/sms/classes/factor.php
```diff
@@ -0,0 +1,358 @@
...
+            return $mform;
+        }
+
+        $mform->addElement(new \tool_mfa\local\form\verification_field());
+        $mform->setType('verificationcode', PARAM_ALPHANUM);
+
+        if (empty($SESSION->mfa_sms_setup_code)) {
+            // They need a code to verify this number.
+            // Lets generate a code and send it in place, we don't need this to get to the DB.
+            $code = random_int(100000, 999999);
```
brendanheywood, line 107: shouldn't this be using the secret manager?

factor/sms/classes/factor.php
```diff
@@ -0,0 +1,358 @@
...
+
+            $this->sms_verification_code($code, $USER->phone2);
+            // Store as string for form comparison.
+            $SESSION->mfa_sms_setup_code = (string) $code;
+            $mform = $this->add_redacted_sent_message($mform);
```
brendanheywood, line 112: move this outside the if?

factor/sms/classes/factor.php
```diff
@@ -0,0 +1,358 @@
...
+        }
+        return $instance->id;
+    }
+
+    /**
+     * This function sends an SMS code to the user based on the instanceid provided.
+     * If rawcode is provided, it will send the raw code through to the phone number in the user profile.
+     *
+     * @param int $secret the secret to send.
+     * @param int|null $rawcode If provided, a raw code to send to the users profile phone no.
```
brendanheywood, line 318: this doesn't match the code

Peterburnett:
> can you add this as a class and move the css to .../styles.css which is magically loaded into the main css file, this way a theme can override it 

Done

Peterburnett:
> move this outside the if? 

Done

Peterburnett:
> shouldn't this be using the secret manager? 
done

factor/sms/lang/en/factor_sms.php
```diff
@@ -0,0 +1,53 @@
...
+$string['settings:aws:key'] = 'Key';
+$string['settings:aws:key_help'] = 'Amazon API key credential.';
+$string['settings:aws:secret'] = 'Secret';
+$string['settings:aws:secret_help'] = 'Amazon API secret credential.';
+$string['settings:aws:region'] = 'Region';
+$string['settings:aws:region_help'] = 'Amazon API gateway region.';
+$string['settings:duration'] = 'Validity duration';
+$string['settings:duration_help'] = 'The period of time that the code is valid.';
+$string['settings:countrycode'] = 'Country number code';
+$string['settings:countrycode_help'] = 'Enter the calling code for the country where user numbers reside. See this link for a list of calling codes: https://en.wikipedia.org/wiki/List_of_country_calling_codes';
```
brendanheywood, line 40: The calling code without the leading + as a default if users do not enter an international number with a + prefix.

See this link for a list of calling codes (+ link)

factor/sms/lang/en/factor_sms.php
```diff
@@ -0,0 +1,53 @@
...
+$string['smssent'] = 'An SMS message containing your verification code was sent to {$a}.';
+$string['smsstring'] = '{$a->code} is your {$a->site} authentication code.
+
+@{$a->url} #{$a->code}';
+$string['info'] = 'Setup a phone number to receive MFA tokens on.';
```
brendanheywood, line 45: Setup a mobile phone number to receive one-time passwords

factor/sms/classes/factor.php
```diff
@@ -0,0 +1,358 @@
...
+     * @return void
+     */
+    private function sms_verification_code($secret, $phonenumber) {
+        global $CFG, $SITE;
+
+        // Here we should get the information, then construct the message.
+        $content = ['site' => $SITE->fullname, 'url' => $CFG->wwwroot, 'code' => $secret];
+        $message = get_string('smsstring', 'factor_sms', $content);
```
brendanheywood, line 326: I think the url can be simplified to just the domain without the http and also without any moodle path suffix. The spec is a little vague on this, but I think less cruft in the message is better

factor/sms/lang/en/factor_sms.php
```diff
@@ -0,0 +1,53 @@
...
+$string['settings:aws:secret'] = 'Secret';
+$string['settings:aws:secret_help'] = 'Amazon API secret credential.';
+$string['settings:aws:region'] = 'Region';
+$string['settings:aws:region_help'] = 'Amazon API gateway region.';
+$string['settings:duration'] = 'Validity duration';
+$string['settings:duration_help'] = 'The period of time that the code is valid.';
+$string['settings:countrycode'] = 'Country number code';
+$string['settings:countrycode_help'] = 'Enter the calling code for the country where user numbers reside. See this link for a list of calling codes: https://en.wikipedia.org/wiki/List_of_country_calling_codes';
+$string['smssent'] = 'An SMS message containing your verification code was sent to {$a}.';
+$string['smsstring'] = '{$a->code} is your {$a->site} authentication code.
```
brendanheywood, line 42: {$a->code} is your {$a->site} one-time security code

factor/sms/lang/en/factor_sms.php
```diff
@@ -0,0 +1,53 @@
...
+
+@{$a->url} #{$a->code}';
+$string['info'] = 'Setup a phone number to receive MFA tokens on.';
+$string['setupfactor'] = 'SMS factor setup';
+$string['summarycondition'] = 'Using an SMS token';
+$string['wrongcode'] = 'Invalid verification code.';
```
brendanheywood, line 48: Invalid security code

factor/sms/lang/en/factor_sms.php
```diff
@@ -0,0 +1,53 @@
...
+$string['settings:duration_help'] = 'The period of time that the code is valid.';
+$string['settings:countrycode'] = 'Country number code';
+$string['settings:countrycode_help'] = 'Enter the calling code for the country where user numbers reside. See this link for a list of calling codes: https://en.wikipedia.org/wiki/List_of_country_calling_codes';
+$string['smssent'] = 'An SMS message containing your verification code was sent to {$a}.';
+$string['smsstring'] = '{$a->code} is your {$a->site} authentication code.
+
+@{$a->url} #{$a->code}';
+$string['info'] = 'Setup a phone number to receive MFA tokens on.';
+$string['setupfactor'] = 'SMS factor setup';
+$string['summarycondition'] = 'Using an SMS token';
```
brendanheywood, line 47: Using an SMS one-time security code

factor/sms/lang/en/factor_sms.php
```diff
@@ -0,0 +1,53 @@
...
+$string['wrongcode'] = 'Invalid verification code.';
+$string['action:revoke'] = 'Revoke SMS factor';
```
brendanheywood, line 49: Should be something more like?

Revoke mobile phone number

factor/sms/lang/en/factor_sms.php
```diff
@@ -0,0 +1,53 @@
...
+$string['awssdkrequired'] = 'The local_aws plugin leveraging the AWS SDK is required to use this factor. Please install local_aws.';
+$string['profilelink'] = 'Click here to edit your profile.';
```
brendanheywood, line 51: remove

factor/sms/classes/factor.php
```diff
@@ -0,0 +1,358 @@
...
+        // Handle phone number submission.
+        if (empty($USER->phone2)) {
+            $DB->set_field('user', 'phone2', $data->phonenumber, ['id' => $USER->id]);
+            // Also update $USER.
+            $USER->phone2 = $data->phonenumber;
+            \core\notification::info(get_string('numberadded', 'factor_sms'));
+            $addurl = new \moodle_url('/admin/tool/mfa/action.php', [
+                'action' => 'setup',
+                'factor' => 'sms',
+                'sesskey' => sesskey(),
```
brendanheywood, line 157: remove sesskey? I don't think we need it here and we don't want it in the url after?

factor/sms/classes/factor.php
```diff
@@ -0,0 +1,358 @@
...
+     * Adds an instance of the factor for a user, from form data.
+     *
+     * {@inheritDoc}
+     */
+    public function setup_user_factor($data) {
+        global $DB, $SESSION, $USER;
+
+        // Handle phone number submission.
+        if (empty($USER->phone2)) {
+            $DB->set_field('user', 'phone2', $data->phonenumber, ['id' => $USER->id]);
```
brendanheywood, line 150: as discussed remove this update

factor/sms/classes/factor.php
Peterburnett: Sesskey is required here, the page clears the URL once loaded, so it never really appears in the url

classes/local/form/verification_field.php
Peterburnett: This will be improved in bonus points round

factor/sms/classes/factor.php
```diff
@@ -0,0 +1,380 @@
...
+     * {@inheritDoc}
+     */
+    public function setup_factor_form_definition($mform) {
+        global $SESSION, $USER;
+        if (empty($USER->phone2) && empty($SESSION->tool_mfa_sms_number)) {
+            $mform->addElement('hidden', 'verificationcode', 0);
+            $mform->setType("verificationcode", PARAM_ALPHANUM);
+
+            // Add field for phone number setup.
+            $mform->addElement('text', 'phonenumber', get_string('addnumber', 'factor_sms'),
```
brendanheywood, line 86: tiny polish bonus, can you add the attr autocomplete="tel" and also inputmode="tel" to key the phone keyboard on mobile

factor/sms/lang/en/factor_sms.php
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @package     factor_sms
+ * @author      Peter Burnett <peterburnett@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+$string['pluginname'] = 'SMS factor';
+$string['privacy:metadata'] = 'The SMS factor plugin does not store any personal data';
+$string['loginsubmit'] = 'Verify Code';
```
brendanheywood, line 28: lower case C on code

factor/sms/lang/en/factor_sms.php
```diff
@@ -0,0 +1,55 @@
...
+
+@{$a->url} #{$a->code}';
+$string['info'] = 'Setup a phone number to receive one-time security codes on.';
+$string['setupfactor'] = 'SMS factor setup';
+$string['summarycondition'] = 'Using an SMS one-time security code';
+$string['wrongcode'] = 'Invalid security code.';
+$string['action:revoke'] = 'Revoke mobile phone number';
+$string['awssdkrequired'] = 'The local_aws plugin leveraging the AWS SDK is required to use this factor. Please install local_aws.';
+$string['addnumber'] = 'Enter mobile phone number';
+$string['phoneplaceholder'] = '0xxxx or +xxxxx';
```
brendanheywood, line 54: I find this slightly misleading, because the lengths are not what people expect. ie my numbers are:

0499 476 733
+61 499 476 733

so I'd expect a placeholder more like:

`04xx xxx xxx   or  +61 4xx xxx xxx`

factor/sms/settings.php
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Settings
+ *
+ * @package     factor_sms
+ * @author      Peter Burnett <peterburnett@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+use totara_core\quickaccessmenu\menu\admin;
```
brendanheywood, line 26: will this bust on moodle?

Peterburnett: That was a stupid IDE autoinclude that I missed, removed

Issue #20, opened 2019-10-28
Have a WebAuthn factor ( Yubikey factor / U2F / Fido / CTAP )


brendanheywood:
https://en.wikipedia.org/wiki/WebAuthn

https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API

Testing demo site:
https://webauthn.io/

Internal WR: https://wrms.catalyst.net.nz/wr.php?request_id=325417

brendanheywood:
Good reading:

https://m.signalvnoise.com/basecamp-now-supports-security-keys-for-two-factor-authentication-thanks-to-webauthn/

Dagefoerde:
here's a rather simple guide on how to implement U2F: https://developers.yubico.com/U2F/Libraries/Using_a_library.html ,
with a U2F server library in PHP at: https://github.com/Yubico/php-u2flib-server (by the Yubikey vendors, but apparently usable with U2F by other vendors too).

Sharing this here because I've been wanting to implement 2FA for Moodle for a long time and had these resources saved for that purpose. Thanks @brendanheywood for creating tool_mfa and for making the necessary changes in core! I love it.

brendanheywood:
you are welcome @Dagefoerde - are you in a position to help with either a pull request or to sponsor the webauth factor?

Dagefoerde:
As I said, I've been wanting to do this as a side project for a while and I might once I find the time. However, right now I can't really make any reliable promises as I am trying to focus on my dissertation.

Dagefoerde:
Working on this in the context of #MootDACH20 now! :)

brendanheywood:
> Working on this in the context of #MootDACH20 now! :)

Cool! :)

arjenlentz:
> Working on this in the context of #MootDACH20 now! :)

Sehr schön, danke @Dagefoerde !

Mfg, Arjen.
(CISO, Catalyst Australia)

nadavkav:
UPDATE: https://android-developers.googleblog.com/2022/10/bringing-passkeys-to-android-and-chrome.html

brendanheywood:
That new spec is very cool - and I think ultimately it should be in core as a first class auth method rather than a second factor. We’d need that to support login to the moodle mobile app working seamlessly with the web login

nadavkav:
@brendanheywood , Is there any Tracker issue about implementing this in core?

brendanheywood:
No but I intend to make one

[...]

PR #230, opened 2020-08-25
Add a U2F factor option

Laur0r:
This plugin lets users define their U2F device. Upon login, it verifies that the defined U2F device is plugged in and the button is pressed.

This plugin is a product of the #MootDACH20 DevCamp (https://moodlemootdach.org/course/view.php?id=13). @Dagefoerde and I have had a lot of fun adding this alternative to tool_mfa! :100: 
Some polishing is required before merging the plugin. However, by creating a pull request early we hope to get some feedback from you.


Dagefoerde:
This will fix #20.

Peterburnett:
Hi all, I should be able to give this a review in the next few days.

factor/u2f/classes/factor.php
```diff
@@ -0,0 +1,251 @@
...
+        global $PAGE, $CFG;
+
+
+
+        // Prepare data for U2F request.
+        $url = parse_url($CFG->wwwroot);
+
+        if($url['scheme'] == "http"){
+            $mform->addElement('html', '<div class="m-element-notification m-element-notification--warning">
+            You cannot use https for u2f authentication.</div></div>');
```
brendanheywood, line 178: typo, http vs https. Also I would use `$OUTPUT->notification('...', \core\output\notification::NOTIFY_ERROR)` instead of the custom html

factor/u2f/classes/factor.php
```diff
@@ -0,0 +1,251 @@
...
+
+    /**
+     * U2F factor implementation.
+     *
+     * {@inheritDoc}
+     */
+    public function possible_states($user) {
+        // U2F can return all states.
+        return array(
+            \tool_mfa\plugininfo\factor::STATE_FAIL,
```
brendanheywood, line 245: I don't think this handles an explicit fail so this should be removed

brendanheywood:
Thanks again @Dagefoerde this is really cool

I've had a very quick glance, I suspect some of the feedback around the planned refactor of the core mfa tables will affect this so I think we should get that sorted and then do a deeper look at this again in a couple weeks

Dagefoerde:
Hi @brendanheywood and @Peterburnett,

Thanks for your comments. 

> I suspect some of the feedback around the planned refactor of the core mfa tables will affect this

I tend to agree. For now, this factor does not store any temporary data (unlike the SMS/Telegram/Email factors). However, I think that it actually should. In its current state, it sends an object regarding details on the request to the client (that's okay, the client needs these details) and trusts the client to send this (hopefully unmodified) object back to the server. This was taken from the demo code of the library, but I don't think that's a good idea. So it would be better to store the request as a "secret" on the server side, replacing this line in particular:

https://github.com/catalyst/moodle-tool_mfa/pull/230/files#diff-75077d9026ee9617c2bf770722b35b85R94

With your planned refactorings in mind, I'll wait until those are in place and will then make this change.

factor/u2f/lang/en/factor_u2f.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @package     factor_u2f
+ * @subpackage  tool_mfa
+ * @author      Jan Dageförde, Laura Troost
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+$string['action:revoke'] = 'Revoke U2F Factor';
+$string['u2f:subject'] = 'Your confirmation code';
+$string['u2f:message'] = 'You are trying to log in to Moodle. Your confirmation code is \'{$a->secret}\'.
```
brendanheywood, line 28: a few copied lang keys not needed

brendanheywood:
hi @Laur0r and @Dagefoerde 

Thanks again, I am very keen to see this one land

The new secret api is in place, but I've also done a bit more reading up on how the yubikeys work under the hood and I'm not sure that this should actually use the secret api after all. As far as I can tell the whole protocol is stateless. We should not need to store anything as there is no communication back to the yubikey but what we should store is the device counter which it looks like you are already doing.

I just tried to give this a proper test and review now but I couldn't get it to work on the latest moodle I got this js error:

```
action.php:557 Uncaught ReferenceError: u2f is not defined
    at action.php:557
    at Object.execCb (require.min.js:5)
    at b.check (require.min.js:5)
    at b.<anonymous> (require.min.js:5)
    at require.min.js:5
    at require.min.js:5
    at each (require.min.js:5)
    at b.emit (require.min.js:5)
    at b.check (require.min.js:5)
    at b.<anonymous> (require.min.js:5)
```

factor/u2f/classes/factor.php
brendanheywood: actually I don't think we should do this check at all. Yes I totally agree that sites should be on https, but this isn't the place to enforce it and there is no technical reason to disallow things here.

factor/u2f/classes/factor.php
```diff
@@ -0,0 +1,251 @@
...
+        ]);
+        $mform->addElement('html', $pressbuttonhtml);
+        $mform->addElement('hidden', 'request', '', ["id" => 'id_request']);
+        $mform->setType('request', PARAM_RAW);
+        $mform->addElement('hidden', 'response_input', '', ['id'=> 'id_response_input']);
+        $mform->setType('response_input', PARAM_RAW);
+        $renderer = $PAGE->get_renderer('core');
+
+        $url = parse_url($CFG->wwwroot);
+        $u2f = new U2F($url['scheme'].'://'.$url['host']);
```
brendanheywood, line 58: its a small edge but multiples moodles on different paths will share the same $appId. If it works I'd just use the raw $CFG->wwwroot here, as far as I can tell $appId is just an arbitrary string and doesn't need to be a url?

Dagefoerde:
> hi @Laur0r and @Dagefoerde
> 
> Thanks again, I am very keen to see this one land
> 
> The new secret api is in place, but I've also done a bit more reading up on how the yubikeys work under the hood and I'm not sure that this should actually use the secret api after all. As far as I can tell the whole protocol is stateless. We should not need to store anything as there is no communication back to the yubikey but what we should store is the device counter which it looks like you are already doing.
> 
> I just tried to give this a proper test and review now but I couldn't get it to work on the latest moodle I got this js error:
> 
> ```
> action.php:557 Uncaught ReferenceError: u2f is not defined
>     at action.php:557
>     at Object.execCb (require.min.js:5)
>     at b.check (require.min.js:5)
>     at b.<anonymous> (require.min.js:5)
>     at require.min.js:5
>     at require.min.js:5
>     at each (require.min.js:5)
>     at b.emit (require.min.js:5)
>     at b.check (require.min.js:5)
>     at b.<anonymous> (require.min.js:5)
> ```

Hi @brendanheywood, thank you, this is much appreciated. I will have a look into whether the secret API has to be used at all. In any case, I am against relying on the request data that the client submits back to us. If we can make this work in a stateless fashion I am all for it. Otherwise, i.e. if we have to store something, I'd rather store it in the secret API.

Which browser are you using? Maybe we have to include the JS library after all. It worked in a recent Firefox without the JS library, though.

factor/u2f/classes/factor.php
Dagefoerde: No, it has to be a URL and it may not include a path. https://developers.yubico.com/U2F/App_ID.html

However, the appID and the request object that are being created here are only valid for a specific request, so I see no issue with multiple moodles sharing the same appID.

brendanheywood:
I'm in chrome on ubuntu:

Your user agent
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.83 Safari/537.36

We have yubikeys nanos and a few other models

brendanheywood:
Also I've been reading up specifically on the OTP USB spec, I believe the u2f spec is a bit more involved so my comments above might be wrong on that front

factor/u2f/classes/factor.php
Dagefoerde: I would definitely agree that this is not the place to perform a check here. However, the thing is that there actually is a technical reason to disallow HTTP here: The U2F protocol refuses to work if the connection is established via HTTP. So this is a hard fail here. And I would keep this check (and notification) in place, primarily because it would make any developer on `http://localhost/` tear their hair out ;)

Removing the custom HTML though, of course.

danmarsden:
We've just landed a webauthn factor in the main release which supports yubikey and a few other different hardware options (but so far we've only tested yubikey and swissbit keys (swissbit sponsored this work.) so I'm closing this PR off for now, but feel free to resubmit a new one if you think there's something missing from our webauthn implementation.

thanks!

PR #234, opened 2020-09-03
Secret manager API

Peterburnett:


classes/secret_manager.php
```diff
@@ -0,0 +1,325 @@
...
+/**
+ * MFA secret management class.
+ *
+ * @package     tool_mfa
+ * @author      Peter Burnett <peterburnett@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+namespace tool_mfa;
```
brendanheywood, line 26: I think this should be in /local/

classes/secret_manager.php
```diff
@@ -0,0 +1,325 @@
...
+     * Inserts the provided secret into the session with a given expiry duration.
+     *
+     * @param string $secret the secret to store
+     * @param integer $expires expiry duration in seconds
+     * @return void
+     */
+    private function add_secret_to_session(string $secret, int $expires) : void {
+        global $SESSION;
+
+        $expirytime = time() + $expires;
```
brendanheywood, line 106: So this is me being devils advocate and just asking

If we want a secret which is tied to the session, do we want to force this to only be available in the session (which is a different question)?


ie I'm thinking ahead to the situation where we have out of band secret approvals, we need to be able to approve a secret without being the active user and having access to the $SESSION, but in this case it still makes sense for the secret to be tied to a single session and not able to be used for other sessions.

In other words should all secrets be stored in the db, but they have an optional sessionid column which is checked as needed?

classes/secret_manager.php
Peterburnett: it currently matches manager.php in namespace. They should both be moved, so I spun that off into https://github.com/catalyst/moodle-tool_mfa/issues/235

classes/local/secret_manager.php
```diff
@@ -0,0 +1,228 @@
...
+    public function __construct(string $factor) {
+        $this->factor = $factor;
+        $this->sessionid = session_id();
+    }
+
+    /**
+     * This function creates or takes a secret, and stores it in the database or session.
+     *
+     * @param integer $expires the length of time the secret is valid. e.g. 1 min = 60
+     * @param boolean $session whether this secret should be stored in the session.
```
brendanheywood, line 48: 'stored in' vs 'linked to'

classes/local/secret_manager.php
```diff
@@ -0,0 +1,228 @@
...
+        return self::NONVALID;
+    }
+
+    /**
+     * Revokes the provided secret code for the user.
+     *
+     * @param string $secret the secret to revoke.
+     * @return void
+     */
+    public function revoke_secret(string $secret) {
```
brendanheywood, line 169: I think this should have $user passed in not $USER, if you revoke from a url in a text or email you might not be logged in yourself (yet) and I don't think we should force people to login then, as they might not be able to if their account is already compromised

classes/local/secret_manager.php
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /**
+     * Deletes any user secrets hanging around in the database.
+     *
+     * @return void
+     */
+    public function cleanup_temp_secrets() {
+        global $DB, $USER;
+        // Session records are autocleaned up.
```
brendanheywood, line 219: moot comment

db/install.xml
```diff
@@ -27,5 +27,25 @@
         <INDEX NAME="lockcounter" UNIQUE="false" FIELDS="userid, factor, lockcounter"/>
       </INDEXES>
     </TABLE>
+    <TABLE NAME="tool_mfa_secrets" COMMENT="Table to store factor secrets">
+      <FIELDS>
+        <FIELD NAME="id" TYPE="int" LENGTH="10" NOTNULL="true" SEQUENCE="true"/>
+        <FIELD NAME="userid" TYPE="int" LENGTH="10" NOTNULL="true" SEQUENCE="false"/>
+        <FIELD NAME="factor" TYPE="char" LENGTH="100" NOTNULL="true" SEQUENCE="false"/>
```
brendanheywood, line 34: should a secret be linked to a factor type or a factor instance id? id seems more correct to me. It might not make a different either way

Peterburnett: I cant think of a use case where the secret would need to be hardlinked to one specific factor instance. The factor type just confirms that this code is indeed for this factor, but it does not necessarily need to be hardlinked to a device for example. If there were 5 email addresses setup for example, all are a valid email code, regardless of the origin address that the code was received at. Happy to change it, but I cant see a use case for it. It is very convenient to Select * where type = email

classes/local/secret_manager.php
Peterburnett: Added $userid as an optional param, defaulting to $USER if null. Makes it easier to use for inband operations, while providing scope for out of band

classes/local/secret_manager.php
Peterburnett: Added as an optional param

Issue #181, opened 2020-04-27
TOTP: Issuer must not contain a colon.


kantgym:
My Moodle Sitename had a ":" in its name. Thus when trying to setup TOTP for a user I received the error  "Issuer must not contain a colon."

Maybe instead of just throwing an error, replacing/escaping the ":" would be better, since ":" is allowed to be put in the sites name.

At least mentioning the requirement not to have ":" in the site-name within the documentation would be nice.

Even better, if the issuer transfered to the TOTP-App would be configurable from the subplugin. Maybe also the format of the account name.

Thanks for this wonderful plugin.

Peterburnett:
Hi @kantgym, thanks for the feedback! This issue appears to arise from the thirdparty libraries restrictions on the characters that can be passed as parameters to the library. As you said, possibly escaping/stripping this character is a solution to this problem.

PR's are welcome to fix this issue if you wish to move this along faster, but this has been added to our backlog.
Thanks

yetanotherbug:
I'd like to expand this bug. This issue also breaks the TOTP factor whenever `$CFG->wwwroot` is set to an URL that contains a port number as part of the netloc (in the sense of RFC 1808).

So wwwroots of the form http://localhost:8080, https://127.0.0.1:8443, https://hurdurnonstandardsslport.com:8888 all cause the TOTP factor to not be configurable because of this code in factor/totp/classes/factor.php:

```php
    public function generate_totp_uri($secret) {
        global $USER, $SITE, $CFG;
        $domain = str_replace('http://', '', str_replace('https://', '', $CFG->wwwroot)); // <----------- bad code, leaves a colon if a port was specified
        $issuer = $SITE->fullname.' '.$domain;
        $totp = TOTP::create($secret);
        $totp->setLabel($USER->username);
        $totp->setIssuer($issuer);
        return $totp->getProvisioningUri();
    }
```

Error invetiably in the user MFA preferences (moodle 3.9 as of v3.9.1-310-g535523f21f8, this plugin as of 2020021401-59-gf6850ee):

```
Exception - Issuer must not contain a colon.

Error code: generalexceptionmessage

Stack trace:

    line 2752 of /admin/tool/mfa/factor/totp/extlib/Assert/Assertion.php: Assert\InvalidArgumentException thrown
    line 1794 of /admin/tool/mfa/factor/totp/extlib/Assert/Assertion.php: call to Assert\Assertion::createException()
[...]

PR #236, opened 2020-09-04
Issue #181: Fix TOTP factor setup for wwwroots and site names containing colons

yetanotherbug:
The parameter to TOTP::setIssuer() must not contain a colon; it's
unsupported by the library. The issuer string includes the site name and
the wwwroot's hostname/domain, so we need to ensure that neither
component can sneak a colon in.

  * In order to include the desired domain in the issuer string, use
    parse_url() to extract the PHP_URL_HOST surgically, without the
    problem of accidentally including port numbers and/or usernames and
    passwords, all of which could be part of the URL.  Standards are in
    agreement that bare host names must not include colons.

    It should be noted then that the "domain" can actually be just an IP
    address, e.g. in the case of "https://127.0.0.1". Rename to hostname.

  * There is no good way to replace colons in human-readable site names
    without maiming the original string in some way.

    We just remove the colon.

    Only the semicolon or the dot is likely to be used with similar
    spacing of preceding and following words (in the English language
    anyway), and replacing the colon with either one would make the
    resulting string even more awkward to read than just removing it.

    According to [1], a colon is the only forbidden character, so we
    should be fine by going this route.

----
[1] https://github.com/Spomky-Labs/otphp/blob/6f1cd6e11a64f8fcd2c8ceb17442f456d310f820/src/ParameterTrait.php#L177

### DIFF
--- a/factor/totp/classes/factor.php
+++ b/factor/totp/classes/factor.php
@@ -55,16 +55,17 @@ class factor extends object_factor_base {
 
     /**
      * Generates TOTP URI for given secret key.
-     * Uses site name, domain and user name to make GA account look like:
-     * "Sitename domain (username)"
+     * Uses site name, hostname and user name to make GA account look like:
+     * "Sitename hostname (username)".
      *
      * @param string $secret
      * @return string
      */
     public function generate_totp_uri($secret) {
         global $USER, $SITE, $CFG;
-        $domain = str_replace('http://', '', str_replace('https://', '', $CFG->wwwroot));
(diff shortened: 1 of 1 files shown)


yetanotherbug:
Disclaimer: I'm neither a PHP programmer nor a Moodle expert and have not traced or tested if Moodle lets invalid hostnames (containing colons in CFG->wwwroot) through or not, for example.

brendanheywood:
thanks @yetanotherbug looks good

PR #251, opened 2020-11-01
Fix target url of 'not enough factors' page

Dagefoerde:
I am starting work on rebasing #230 / #231 PRs on the new codebase. 

As a first step, I factored out  a change to `renderer.php` as it does not logically relate to the two new factors. Without the change, it is impossible to log out and re-try login if not enough factors could be supplied. In particular, this is a problem when testing a new factor implementation ;) 

IIRC this change was originally done by @Laur0r, so I attributed the commit to her.


brendanheywood:
thanks @Dagefoerde this got fixed somewhere along the way already so closing this

PR #252, opened 2020-11-01
Add a Telegram factor option

Dagefoerde:
[Replaces #231]

This plugin lets users define their Telegram ID. Upon login, it generates a six-digit code and sends it to the defined Telegram ID. The plugin is a product of the #MootDACH20 DevCamp (https://moodlemootdach.org/course/view.php?id=13). @Laur0r and I have had a lot of fun adding this alternative to tool_mfa! :100: Now that DevCamp is over, and some time has passed, I finally got around to do some more work. As there is a new secret manager API, I have created the plugin from scratch.

Still needs testing and a bit of improvement in the admin UI, so WIP.

### DIFF
--- a/renderer.php
+++ b/renderer.php
@@ -224,7 +224,7 @@ class tool_mfa_renderer extends plugin_renderer_base {
         $return = $this->output->notification($notification, 'notifyerror');
 
         // Logout button.
-        $url = new \moodle_url('\admin\tool\mfa\auth.php', ['logout' => 1]);
+        $url = new \moodle_url('/admin/tool/mfa/auth.php', ['logout' => 1]);
         $btn = new \single_button($url, get_string('logout'), 'post', true);
         $return .= $this->render($btn);
 
--- a/factor/telegram/classes/factor.php
+++ b/factor/telegram/classes/factor.php
@@ -48,8 +48,8 @@ class factor extends object_factor_base {
      * {@inheritDoc}
      */
     public function login_form_definition_after_data($mform) {
-        $instanceid = $this->generate_and_send_code();
-        $mform = $this->add_redacted_sent_message($mform, $instanceid);
+        $recipient = $this->generate_and_send_code();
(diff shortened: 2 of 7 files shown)


Dagefoerde:
I have added help texts for admins and users, thus improving the UI overall. Also I have tested the plugin's functionality and believe that, in my opinion, it is production-ready now. Happy to hear your reviews.

renderer.php
```diff
@@ -224,7 +224,7 @@ public function not_enough_factors() {
         $return = $this->output->notification($notification, 'notifyerror');
 
         // Logout button.
-        $url = new \moodle_url('\admin\tool\mfa\auth.php', ['logout' => 1]);
+        $url = new \moodle_url('/admin/tool/mfa/auth.php', ['logout' => 1]);
```
brendanheywood, line 227: I've fixed bug this in master please remove it from here - thanks nice catch :)

brendanheywood:
hi @Dagefoerde this is looking pretty good. Setting up the bot as admin was smooth, and I really like the info bot which shows you your id. I did have some issues:


1) I did run into a weird loop where I entered my setup code and then it asked for it again over and over. I revoked it and tried a second time and I got this error:

```
! ) Notice: Undefined property: stdClass::$telegramid in /var/www/moodle.local/admin/tool/mfa/factor/telegram/classes/factor.php on line 274
--


1 | 0.0000 | 365664 | {main}( ) | .../action.php:0
2 | 0.9307 | 7629320 | factor_telegram\factor->setup_user_factor( ) | .../action.php:93
```

2) When you try to setup a telegram user, I could not get the username to work but I could use the telegram id. But if you try to register something which doesn't work, and you go back and start over, then it remembers the username or id from before and you can't enter a new one. I think both of these issues stem from using the session variable, I think this should be stateless and part of the form definition. ie the form has one element for the userid, you save it and it fails validation, re-renders the form and now adds the second form in based on the partial form $data rather then session. In theory you could register two telegram ids at the same this way and they won't clash.


After that gets fixed I think this is good to land, good stuff

PR #259, opened 2021-01-12
Issue #258: Added bulk action interface for resetting factors

### LINKED ISSUE #258: Create a bulk action interface for resetting user factors

Peterburnett:
Closes #258

### DIFF
--- a/reset_factor.php
+++ b/reset_factor.php
@@ -27,27 +27,53 @@ require_once(__DIR__ . '/../../../config.php');
 require_once($CFG->libdir . '/adminlib.php');
 admin_externalpage_setup('tool_mfa_resetfactor');
 
+if (!empty($SESSION->bulk_users)) {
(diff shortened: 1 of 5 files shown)


reset_factor.php
```diff
@@ -27,27 +27,53 @@
 require_once($CFG->libdir . '/adminlib.php');
 admin_externalpage_setup('tool_mfa_resetfactor');
 
+if (!empty($SESSION->bulk_users)) {
+    $bulk = true;
+    $users = array_map(function($user) {
+        return \core_user::get_user($user);
```
brendanheywood, line 33: This would be very heavy on the memory in one big hit. How about if if you left this array as just the ids, and did the user record lookup incrementally in the loop below so you only held one user in memory at a time? The loop can figure out if it is given just an id or a full object

PR #262, opened 2021-01-25
Issue #261: Fixed issue where special roles not respected

### LINKED ISSUE #261: Role factor does not respect "special" roles (authenticated user)

Peterburnett:
Closes #261


brendanheywood:
good to go when green

PR #269, opened 2021-02-05
Fixed ordering bug with auth method reordering

Peterburnett:


factor/auth/version.php
```diff
@@ -26,7 +26,7 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version      = 2019102400;      // The current plugin version (Date: YYYYMMDDXX).
+$plugin->version      = 2021020500;      // The current plugin version (Date: YYYYMMDDXX).
 $plugin->requires     = 2017051500.00;   // Support back to 3.3 - Totara 12. Patches required.
 $plugin->component    = 'factor_auth';
 $plugin->release      = 'v0.1';
```
brendanheywood, line 32: unrelated but can you bump this in lockstep with version

PR #278, opened 2021-02-17
Added settings around showing totp link #277

brendanheywood:
#277


Peterburnett:
Just needs a version bump to the TOTP factor so the new setting gets a value and strings are updated

PR #281, opened 2021-02-18
Issue #280: Added debug output to the auth fail screen

### LINKED ISSUE #280: Show debug mode on the Failed auth page if enabled
It will not give significant info on passing factors, but allows viewing the config and seeing the ordering and flow of factors.

Peterburnett:
Closes #280

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -186,6 +186,9 @@ class manager {
         $renderer = $PAGE->get_renderer('tool_mfa');
 
         echo $renderer->header();
+        if (get_config('tool_mfa', 'debugmode')) {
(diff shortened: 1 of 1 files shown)


classes/manager.php
```diff
@@ -186,6 +186,9 @@ public static function cannot_login() {
         $renderer = $PAGE->get_renderer('tool_mfa');
 
         echo $renderer->header();
+        if (get_config('tool_mfa', 'debugmode')) {
+            self::display_debug_notification();
```
brendanheywood, line 190: hhmm, it makes me thing display_debug_notification should return the html?

PR #302, opened 2021-06-27
Replace all non-alphanumeric characters in the site shortname for sender ID

### LINKED ISSUE #301: Error with SMS factor when site shortname contains non alphanumeric characters
On the AWS docs: https://docs.aws.amazon.com/sns/latest/api/API_SetSMSAttributes.html#API_SetSMSAttributes_RequestParameters
`DefaultSenderID – A string, .... The sender ID can be 1 - 11 alphanumeric characters, and it must contain at least one letter. `

In the tool_mfa code:
https://github.com/catalyst/moodle-tool_mfa/blob/master/factor/sms/classes/local/smsgateway/aws_sns.php#L73
The sender ID is set to the site shortname, which doesn't limit the ability to input non-alphanumeric characters.
Because of this when we have a site shortname containing a character such as '&' we get this error:
```
Error executing "SetSMSAttributes" on "https://sns.ap-southeast-2.amazonaws.com"; AWS HTTP error: Client error: POST https://sns.ap-southeast-2.amazonaws.com resulted in a 400 Bad Request response:
InvalidPara (truncated...)
 InvalidParameter (client): Invalid parameter:  - Sender
```

this could be fixed by doing:
`$senderid = preg_replace("/[^A-Za-z0-9]/", '', trim($senderid));`
on this line: https://github.com/catalyst/moodle-tool_mfa/blob/master/factor/sms/classes/local/smsgateway/aws_sns.php#L65

This fix would also remove the spaces that the current str_replace does.

alexmorrisnz:
Fixes #301

### DIFF
--- a/factor/sms/classes/local/smsgateway/aws_sns.php
+++ b/factor/sms/classes/local/smsgateway/aws_sns.php
@@ -61,8 +61,8 @@ class aws_sns implements gateway_interface {
 
         // Setup the sender information.
         $senderid = $SITE->shortname;
-        // Remove spaces from ID.
-        $senderid = str_replace(' ', '', (trim($senderid)));
+        // Remove spaces and non-alphanumeric characters from ID.
+        $senderid = preg_replace("/[^A-Za-z0-9]/", '', trim($senderid));
         // We have to truncate the senderID to 11 chars.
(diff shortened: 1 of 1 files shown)


alexmorrisnz:
The failing check looks unrelated.

PR #305, opened 2021-08-12
factor_sms: Add new Modica gateway

alexmorrisnz:


factor/sms/classes/local/smsgateway/modica.php
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * This function internationalises a number to E.164 standard.
+     * https://46elks.com/kb/e164
+     *
+     * @param string $phonenumber the phone number to format.
+     * @return string the formatted phone number.
+     */
+    private function format_number(string $phonenumber): string {
```
alexmorrisnz, line 77: Will move this into a helper class

factor/sms/classes/local/smsgateway/aws_sns.php
```diff
@@ -28,6 +28,11 @@
 defined('MOODLE_INTERNAL') || die();
 
 require_once($CFG->dirroot . '/local/aws/sdk/aws-autoloader.php');
+
+use admin_setting_configcheckbox;
```
Peterburnett, line 32: Should probably just  require lib/adminlib.php

factor/sms/classes/local/smsgateway/modica.php
```diff
@@ -0,0 +1,111 @@
...
+     *
+     * {@inheritDoc}
+     */
+    public function send_sms_message(string $messagecontent, string $phonenumber): bool {
+        global $CFG, $SITE;
+
+        $config = get_config('factor_sms');
+
+        $params = [
+            'http' => 'https://api.modicagroup.com/rest/gateway',
```
Peterburnett, line 46: maybe an admin setting? likely not needed just a thought

alexmorrisnz: I don't think it will change but I've added it anyway

alexmorrisnz:
rebased with the latest master commits

alexmorrisnz:
Hi @brendanheywood @Peterburnett I've got it passing tests now for pgsql
The mariadb failing tests are unrelated and are also happening on the main branch

factor/sms/classes/helper.php
```diff
@@ -0,0 +1,59 @@
...
+ * Helper class for shared sms gateway functions
+ *
+ * @package     factor_sms
+ * @author      Alex Morris <alex.morris@catalyst.net.nz>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+namespace factor_sms;
+
```
Peterburnett, line 27: double whitepsace

factor/sms/classes/local/smsgateway/aws_sns.php
```diff
@@ -27,8 +27,9 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-require_once($CFG->dirroot . '/local/aws/sdk/aws-autoloader.php');
-use Aws\Sns\SnsClient;
+if (file_exists($CFG->dirroot . '/local/aws/version.php')) {
```
Peterburnett, line 30: I dont think the check for the AWS API should be here. You likely shouldnt be able to reach this code path at all if it isnt installed, should probably not even be an option at the settings level.

if we dont load the SDK, it will just exception when we try to use it anyway

factor/sms/classes/local/smsgateway/aws_sns.php
```diff
@@ -86,31 +87,46 @@ public function send_sms_message(string $messagecontent, string $phonenumber) :
...
+                'ap-southeast-2'));
         }
+    }
 
-        // Prepend country code.
-        $countrycode = get_config('factor_sms', 'countrycode');
-        $phonenumber = '+' . $countrycode . $phonenumber;
-
-        return $phonenumber;
+    public function is_gateway_enabled() : bool {
```
Peterburnett, line 125: Same here, I think this could potentially be static, I cant imagine an object level state check being needed for this.

Issue #307, opened 2021-09-08
Each factor should have a method to declare if it is lockable or not


brendanheywood:
ie if you pass or fail a factor such as 'do you have this role', then you should never get locked out because of this.

In practice you don't, but it's making a redundant sql call  which we can eliminate for those that don't even need a concept of locking

SELECT MAX(lockcounter) FROM mdl_tool_mfa WHERE userid = $1 AND factor = $2

Declare non locking for all the 'user filtering' factors as a starting point, in fact it might be better to opt in rather than opt out. 

https://github.com/catalyst/moodle-tool_mfa#user-filtering-factors

Peter says this should be 1-1 with 'input factors' - I'm (Brendan) wondering if its worth still having a 'supports_locking' method but its in the base and just checks if its an input factor and pretty well everything just inherits it

PR #308, opened 2021-09-09
Lockcount performance improvements

brendanheywood:
https://github.com/catalyst/moodle-tool_mfa/issues/307


classes/local/factor/object_factor_base.php
```diff
@@ -59,21 +59,39 @@ public function __construct($name) {
...
         $this->secretmanager = new \tool_mfa\local\secret_manager($this->name);
+    }
+
+    /**
+     * This loads the locked state from the DB
+     *
+     * Base class implementation.
+     *
+     */
+    public function check_locked_state() {
```
Peterburnett, line 70: As this doesnt really return anything, its name might be better as process_locking_state or similar?

PR #310, opened 2021-09-17
factor_sms: Add SMS sent event

alexmorrisnz:


factor/sms/classes/local/smsgateway/aws_sns.php
```diff
@@ -78,10 +80,25 @@ public function send_sms_message(string $messagecontent, string $phonenumber) :
...
                 'PhoneNumber' => $phonenumber,
             ]);
+
+            $data = array(
+                'relateduserid' => null,
+                'context' => \context_user::instance($USER->id),
+                'other' => array(
+                    'userid' => $USER->id,
+                    'debug' => array(
+                        'phonenumber' => $phonenumber,
```
Peterburnett, line 94: I think we should likely redact the phone number. Logs are generally sensitive but I dont think there is a reason to keep it in, and its 1 less piece of sensitive information.

https://github.com/catalyst/moodle-tool_mfa/blob/01635da474c4b863374de215dbccaacc2caacd32/factor/sms/classes/factor.php#L235 This could be moved to a function in the helper. Not mission critical however

factor/sms/classes/event/sms_sent.php
```diff
@@ -0,0 +1,44 @@
...
+
+class sms_sent extends \core\event\base {
+
+    protected function init() {
+        $this->data['crud'] = 'r';
+        $this->data['edulevel'] = self::LEVEL_OTHER;
+    }
+
+    public function get_description() {
+        return "The user with id '{$this->other['userid']}' had an SMS message sent to them. <br> Information: {$this->other['debug']}";
```
Peterburnett, line 38: maybe this should be a tiny bit more specific to MFA, eg `had a verification code sent to the via SMS` or similar

PR #311, opened 2021-09-20
Add forced MFA setup to gracemode

Peterburnett:


factor/grace/classes/factor.php
```diff
@@ -84,8 +84,8 @@ public function has_input() {
      *
      * {@inheritDoc}
      */
-    public function get_state() {
-        global $USER;
+    public function get_state($noredir = false) {
```
keevan, line 87: I'm okay with params and defaults, but negative sounding booleans was never something I was fond about. Could this be changed to something like `redirect = true` instead?

factor/grace/classes/factor.php
```diff
@@ -111,9 +111,28 @@ public function get_state() {
...
+                if (time() > $starttime + $duration) {
+                    // If gracemode would have given points, but now doesnt,
+                    // Jump out of the loop and force a factor setup.
+                    // We will return once there is a setup, or the user tries to leave.
+                    if (get_config('factor_grace', 'forcesetup') && !$noredir) {
+                        $redir = true;
+                        $factorurls = \tool_mfa\manager::get_no_redirect_urls();
+                        foreach ($factorurls as $factorurl) {
+                            if ($factorurl->compare($PAGE->url)) {
+                                $redir = false;
```
keevan, line 123: Once you rename the above to `$redirect = true` this probably doesn't need to exist, and you'd be disabling redirect in lines 123 only. 

If it makes more sense from a caller perspective, perhaps the $redirect param could be $redirectable instead, as redirects aren't always performed.

factor/grace/classes/factor.php
```diff
@@ -111,9 +111,28 @@ public function get_state() {
...
+                        }
+
+                        if ($redir) {
+                            redirect(new \moodle_url('/admin/tool/mfa/user_preferences.php'),
+                                get_string('redirectsetup', 'factor_grace'));
+                        }
+                    }
+                    return \tool_mfa\plugininfo\factor::STATE_NEUTRAL;
+                } else {
+                    \tool_mfa\plugininfo\factor::STATE_PASS;
```
keevan, line 134: Missing `return` keyword

factor/grace/classes/factor.php
```diff
@@ -184,4 +203,25 @@ public function possible_states($user) {
             \tool_mfa\plugininfo\factor::STATE_NEUTRAL
         );
     }
+
+    /**
+     * Grace factor implementation.
+     *
+     * If grace period should redirect at end, make this a no-redirect url.
+     *
+     * @return void
```
keevan, line 212: Seems to return an `array` all the time.

factor/grace/classes/factor.php
```diff
@@ -184,4 +203,25 @@ public function possible_states($user) {
...
+     */
+    public function get_no_redirect_urls() {
+        $redirect = get_config('factor_grace', 'forcesetup');
+
+        if ($this->get_state(true) === \tool_mfa\plugininfo\factor::STATE_NEUTRAL && $redirect) {
```
keevan, line 217: Micro optimisation but I'd put simpler boolean checks first in a multi-conditional

factor/grace/classes/factor.php
```diff
@@ -84,8 +84,8 @@ public function has_input() {
      *
```
keevan, line 84: Missing param descriptor in phpdocs

factor/grace/classes/factor.php
keevan: Also how is this not something PHP linting/CS/CI doesn't pick up. A no-op line should be flagged imo.

PR #323, opened 2022-01-19
Issue #48: Token factor

### LINKED ISSUE #48: Have a 'trust this device' option on auth page
This might share some code with a 'recent devices' factor:

https://github.com/catalyst/moodle-tool_mfa/issues/15

Broadly we'd record some sort of fingerprint for a device, maybe a long lived cookie, + the user agent, and possible also combined with a ip address, and allow people to trust it.

image

The admins might be able to say things like only allow people to trust devices on a certain network.

Peterburnett:
Closes #48

### DIFF
--- a/factor/token/classes/factor.php
+++ b/factor/token/classes/factor.php
@@ -26,11 +26,21 @@
 namespace factor_token;
 
 use tool_mfa\local\factor\object_factor_base;
+use tool_mfa\local\secret_manager;
 
 class factor extends object_factor_base {
 
     /**
-     * token implementation.
+     * Token implementation.
+     *
+     * {@inheritDoc}
+     */
+    public function has_input() {
+        return false;
+    }
+
+    /**
+     * Token implementation.
      * This factor is a singleton, return single instance.
      *
      * {@inheritDoc}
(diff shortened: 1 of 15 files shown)


factor/token/classes/factor.php
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * Token Implementation.
+     * We can't get_state like the parent here or it will recurse forever.
+     *
+     * @param string $state
+     * @return void
+     */
+    public function set_state($state) {
+        global $SESSION;
+        $property = 'factor_'.$this->name;
```
keevan, line 122:
```suggestion
        $property = 'factor_' . $this->name;
```

factor/token/classes/privacy/provider.php
```diff
@@ -0,0 +1,46 @@
...
+
+namespace factor_token\privacy;
+
+use core_privacy\local\metadata\null_provider;
+use core_privacy\local\legacy_polyfill;
+
+/**
+ * Class provider
+ * @package factor_token\privacy
+ */
```
keevan, line 33: Same as previous

factor/token/tests/factor_test.php
```diff
@@ -0,0 +1,71 @@
...
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Tests for MFA manager class.
+ *
+ * @package     tool_mfa
+ * @author      Peter Burnett <peterburnett@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
```
keevan, line 24: Same as previous (swapping)

factor/token/lang/en/factor_token.php
```diff
@@ -0,0 +1,35 @@
...
+ * @author      Peter Burnett <peterburnett@catalyst-au.net>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+$string['pluginname'] = 'Trust this device';
+
+$string['event:token_created'] = 'MFA token created.';
+$string['form:trust'] = 'Trust this device for {$a}.';
+$string['privacy:metadata'] = 'The  token factor plugin does not store any personal data';
```
keevan, line 30: Double space typo here, and should it end in a full stop?

factor/token/classes/factor.php
```diff
@@ -0,0 +1,251 @@
...
+
+        list($expirytime, $expiry) = $this->calculate_expiry_time();
+
+        // Store this secret in the database.
+        $secretmanager = new secret_manager($this->name);
+        $secret = base64_encode(random_bytes(256));
+        $secretmanager->create_secret($expiry, false, $secret);
+
+        // All the prep is now done, we can set this cookie.
+        setcookie($cookie, $secret, $expirytime, '/admin/tool/mfa', '', false, true);
```
keevan, line 204: Correct me if I'm wrong but won't this just store and verify the MFA expiry time in a cookie? i.e. If someone were to provide a valid MFA auth on their account, and edit their cookie name to pass this condition for another userid?

I don't see the userid being used as part of the verification/encrypted payload so I'm curious. That said I'm only saying what I'm seeing from the code.

factor/token/classes/factor.php
```diff
@@ -0,0 +1,251 @@
...
+        if (!property_exists($SESSION, 'tool_mfa_factor_token')) {
+            return;
+        }
+        $settoken = $SESSION->tool_mfa_factor_token;
+        if (!$settoken) {
+            return;
+        }
+
+        $userid = $USER->id;
+        $cookie = 'MFA_TOKEN_' . $userid;
```
keevan, line 194:
```suggestion
        $cookie = 'MFA_TOKEN_' . $USER->id;
```

Should just use the variable directly since it's only used once in this function

factor/token/classes/factor.php
```diff
@@ -0,0 +1,251 @@
...
+        // Check if there was a previous locked status to return.
+        $state = parent::get_state();
+        if ($state === \tool_mfa\plugininfo\factor::STATE_LOCKED) {
+            return \tool_mfa\plugininfo\factor::STATE_LOCKED;
+        }
+
+        // Check cookie Exists.
+        if (!NO_MOODLE_COOKIES) {
+            $userid = $USER->id;
+            $cookie = 'MFA_TOKEN_' . $userid;
```
keevan, line 87:
```suggestion
            $cookie = 'MFA_TOKEN_' . $USER->id;
```

Should just use the variable directly since it's only used once in this function

factor/token/classes/factor.php
```diff
@@ -0,0 +1,251 @@
...
+
+            if (!empty($_COOKIE[$cookie])) {
+                $token = $_COOKIE[$cookie];
+            } else {
+                return \tool_mfa\plugininfo\factor::STATE_NEUTRAL;
+            }
+        } else {
+            return \tool_mfa\plugininfo\factor::STATE_NEUTRAL;
+        }
```
keevan, line 96: I think you can de-nest this block, e.g. put the else in it's own condition and early return
```
    if (NO_MOODLE_COOKIES) {
        return \tool_mfa\plugininfo\factor::STATE_NEUTRAL;
    }
```

Then later:
```
            if (empty($_COOKIE[$cookie])) {
                return \tool_mfa\plugininfo\factor::STATE_NEUTRAL;
            }
            $token = $_COOKIE[$cookie];

```

keevan (commented):
Hi Peter, thanks for the PR I've only done a first pass to check the code

factor/token/tests/factor_test.php
```diff
@@ -0,0 +1,71 @@
...
+
+namespace factor_token\tests;
+
+defined('MOODLE_INTERNAL') || die();
+
+class factor_token_test extends \advanced_testcase {
```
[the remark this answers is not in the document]
Peterburnett: From some quick checking this will break backwards compatability. Leaving in place for now

factor/token/classes/factor.php
Peterburnett: the Secret manager has implicit user checks. the userid on the token is just for deduplication of tokens if a user controls multiple lms accounts on the same browser

PR #328, opened 2022-02-04
Issue #327: Allow grace to define an ignorelist

### LINKED ISSUE #327: Allow gracemode to have an ignorelist of factors
Recently a client issue was encountered where new users were unable to enter the site, as email factor was enabled, however there were significant delivery issues with the email outside of our control.

A nice workaround to allow users to enter the site and configure other factors such as TOTP would be an ignorelist of factors for gracemode, that do not contribute to the points counter here: https://github.com/catalyst/moodle-tool_mfa/blob/master/factor/grace/classes/factor.php#L95

Peterburnett:
Closes #327

### DIFF
--- a/factor/grace/classes/factor.php
+++ b/factor/grace/classes/factor.php
@@ -89,16 +89,14 @@ class factor extends object_factor_base {
         $record = reset($records);
 
         // First check if user has any other input or setup factors active.
-        $factors = \tool_mfa\plugininfo\factor::get_active_user_factor_types();
+        $factors = $this->get_affecting_factors();
         $total = 0;
         foreach ($factors as $factor) {
-            if ($factor->has_input() || $factor->has_setup()) {
-                $total += $factor->get_weight();
-                // If we have hit 100 total, then we know it is possible to auth with the current setup.
-                // Gracemode should no longer give points.
-                if ($total >= 100) {
-                    return \tool_mfa\plugininfo\factor::STATE_NEUTRAL;
-                }
+            $total += $factor->get_weight();
+            // If we have hit 100 total, then we know it is possible to auth with the current setup.
+            // Gracemode should no longer give points.
+            if ($total >= 100) {
+                return \tool_mfa\plugininfo\factor::STATE_NEUTRAL;
             }
         }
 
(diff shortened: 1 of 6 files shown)


factor/grace/lang/en/factor_grace.php
```diff
@@ -29,6 +29,8 @@
 $string['settings:forcesetup_help'] = 'Forces a user to the preferences page to setup MFA when the gracemode period expires. If set to off, users will be unable to authenticate when the grace period expires.';
 $string['settings:graceperiod'] = 'Grace period';
 $string['settings:graceperiod_help'] = 'Period of time when users can access Moodle without configured and enabled factors';
+$string['settings:ignorelist'] = 'Ignored factors';
+$string['settings:ignorelist_help'] = 'Any factors here will not be counted by grace when deciding whether to give points.';
```
dmitriim, line 33: Sorry, but I can't understand what it is really means for the users :( Maybe add an example?

factor/grace/settings.php
```diff
@@ -37,6 +37,18 @@
...
 
-$name = new lang_string('settings:graceperiod', 'factor_grace');
-$description = new lang_string('settings:graceperiod_help', 'factor_grace');
-$settings->add(new admin_setting_configduration('factor_grace/graceperiod', $name, $description, '604800'));
+$settings->add(new admin_setting_configduration('factor_grace/graceperiod',
+    new lang_string('settings:graceperiod', 'factor_grace'),
+    new lang_string('settings:graceperiod_help', 'factor_grace'), '604800'));
+
+// Get factors that gracemode cares about.
+$activefactors = \tool_mfa\plugininfo\factor::get_enabled_factors();
```
dmitriim, line 45: will the setting screw up when we enable/disable factors and then save it? Maybe better to use all factors to be safe? and then check if it's enabled later (if required of course)...

factor/grace/settings.php
```diff
@@ -37,6 +37,18 @@
...
+$active = [];
+foreach ($activefactors as $factor) {
+    if ($factor->has_setup() || $factor->has_input()) {
```
dmitriim, line 48: why do we do it here?

dmitriim:
would be good to add any kind of auto tests if possible and makes sense to add them. I'm ok to abstract a new code to a method and then test just it for this PR.

factor/grace/settings.php
Peterburnett: Gracemode only cares about factors that a user can input or affect. Filtering here makes the list a lot shorter and less cluttered.

factor/grace/classes/factor.php
```diff
@@ -236,4 +234,41 @@ public function get_no_redirect_urls() {
...
+    /**
+     * Returns a list of factor objects that can affect gracemode giving points.
+     *
+     * @return array
+     */
+    public function get_all_affecting_factors(): array {
+        // Check if user has any other input or setup factors active.
+        $factors = \tool_mfa\plugininfo\factor::get_factors();
+        $factors = array_filter($factors, function($el) {
+            return $el->has_input() || $el->has_setup();
```
dmitriim, line 247: @Peterburnett  why do we need this check to filer out factors?

PR #334, opened 2022-03-14
Fix redirect loop explosions when running moodle on location != /

### LINKED ISSUE #333: Redirect loop explosions when running moodle on location != /
When running Moodle with something like `$CFG->wwwroot = 'https://example.com/moodle';` and tool_mfa enabled, any attempt to login results in "Unsupported redirect detected, script execution terminated", similar to #148.

accerqueira:
In order to fix #333, I think `$PAGE->set_url('/admin/tool/mfa/auth.php');` must be executed before `require_login(null, false);` on https://github.com/catalyst/moodle-tool_mfa/blob/master/auth.php#L35, so that https://github.com/catalyst/moodle-tool_mfa/blob/master/classes/manager.php#L598 works properly, when Moodle is not setup directly on location `/`.

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -28,14 +28,14 @@ require_once($CFG->libdir.'/adminlib.php');
 
 use tool_mfa\local\form\login_form;
 
-require_login(null, false);
-
 $context = context_user::instance($USER->id);
 $PAGE->set_context($context);
 $PAGE->set_url('/admin/tool/mfa/auth.php');
 $PAGE->set_pagelayout('secure');
 $pagetitle = $SITE->shortname.': '.get_string('mfa', 'tool_mfa');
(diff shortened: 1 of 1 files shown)


Peterburnett:
This fix seems fine, but I think there is another issue with the alternate detection using $FULLME, and with the ongoing comparisons using URL_MATCH_BASE. Are you able to do a little more debugging on the redirect check and see if you can nail down that issue and amend the patch? Id wager auth isnt the only page with the issue

danmarsden:
I'm closing this PR - it doesn't look right to me to acess $USER after a require_login call - we should be checking for login before then and the solution that Peter suggests above sounds like a better approach - feel free to submit another PR in future if you have a better solution in future - thanks!

PR #337, opened 2022-03-29
TOTP setup page UX tighten up

Peterburnett:


classes/local/form/verification_field.php
```diff
@@ -43,11 +43,16 @@ public function __construct($attributes=null, $submit = true) {
         }
 
         $attributes['autocomplete'] = 'one-time-code';
-        $attributes['autofocus'] = 'autofocus';
         $attributes['inputmode'] = 'numeric';
         $attributes['pattern'] = '[0-9]*';
         $attributes['class'] = 'tool-mfa-verification-code';
 
+        // If we aren't autosubmitting, this might be part of a larger form such as for setup.
```
keevan, line 50: What did you mean by _autosubmitting_ . This comment may need to be cleaned up slightly if I'm reading that wrong. 

Perhaps:
```suggestion
        // If this is part of a larger form, such as for setup, then we shouldn't apply autofocus.
```

factor/totp/classes/factor.php
```diff
@@ -42,6 +42,7 @@
 require_once(__DIR__.'/../extlib/ParagonIE/ConstantTime/Binary.php');
 require_once(__DIR__.'/../extlib/ParagonIE/ConstantTime/Base32.php');
 
+use html_writer;
```
keevan, line 45: I think it would be better to either keep this and remove the `\` from all the code below, or - my preference - keep using `\html_writer` and remove this statement instead. 

There is one instance (which I'll point out) which is using the non-backslash form, and I think this use statement is a bit redundant.

factor/totp/classes/factor.php
```diff
@@ -163,8 +169,23 @@ public function setup_factor_form_definition_after_data($mform) {
...
 
         $html = \html_writer::table($manualtable);
-        $mform->addElement('static', 'enter', get_string('setupfactor:enter', 'factor_totp'), $html);
-        $mform->addHelpButton('enter', 'setupfactor:enter', 'factor_totp');
+        $html = \html_writer::tag('p', get_string('setupfactor:enter', 'factor_totp')) . $html;
+        // Wrap the table in a couple of divs to be controlled via bootstrap.
+        $html = \html_writer::div($html, 'card card-body', ['style' => 'padding-left: 0 !important;']);
+        $html = \html_writer::div($html, 'collapse', ['id' => 'collapseManualAttributes']);
+
+        $togglelink = html_writer::tag('btn', get_string('setupfactor:scanfail', 'factor_totp'), [
```
keevan, line 177: `html_writer` (please see previous comment)

keevan (commented):
Few small comments, mainly styling.

factor/totp/classes/factor.php
Peterburnett: Stupid IDE trying to be helpful. Yeah I generally just use it backspaced, must have missed one. Nice catch

classes/local/form/verification_field.php
```diff
@@ -34,7 +34,15 @@ class verification_field extends \MoodleQuickForm_text {
...
-    public function __construct($attributes=null, $submit = true) {
+    /**
+     * Verification field is a text entry box that features some useful extras.
+     *
+     * Contains JS to autosubmit the auth page when entered, as well as additional styling.
+     *
+     * @param array $attributes
+     * @param boolean $auth is this field used in auth.php?
+     */
+    public function __construct($attributes=null, $auth = true) {
```
keevan, line 45: Inconsistent spacing detected :zany_face:

PR #338, opened 2022-03-30
Convert grace warning to HTML aware optional content

Peterburnett:


classes/local/form/verification_field.php
```diff
@@ -40,9 +40,9 @@ class verification_field extends \MoodleQuickForm_text {
      * Contains JS to autosubmit the auth page when entered, as well as additional styling.
      *
      * @param array $attributes
-     * @param boolean $auth is this field used in auth.php?
+     * @param boolean $auth is this field used in auth.php
```
keevan, line 43: I still feel like it's unclear. Perhaps this suggestion would be slightly better?
```suggestion
     * @param boolean $auth whether or not the auth page is being entered
```

PR #345, opened 2022-05-02
Updated date display for lastverified to handle 0 correctly

Peterburnett:


renderer.php
```diff
@@ -162,7 +162,9 @@ public function active_factors() {
                 $timecreated  = $userfactor->timecreated == '-' ? '-'
                     : userdate($userfactor->timecreated,  get_string('strftimedatetime'));
                 $lastverified = $userfactor->lastverified;
-                if ($lastverified != '-') {
+                if ($lastverified == 0) {
+                    $lastverified = '-';
```
brendanheywood, line 166: would this be better as a string saying 'Not yet verified' ?

Peterburnett: Split this to a new issue here:
https://github.com/catalyst/moodle-tool_mfa/issues/346

Issue #343, opened 2022-04-25
Totara 13+ Compatibility


LiamKearn:
Sort of a duplicate of #339.

Started to utilise this plugin on a Totara 15 instance and noticed the [time based factor's form](https://github.com/catalyst/moodle-tool_mfa/blob/642cb20ca95c4d94d054494cabb3fd472a39c44f/factor/totp/classes/factor.php#L146) didn't display a few elements as expected. After a quick look I'm fairly certain that changes in Totara's codebase namely the following from `lib/upgrade.txt` (introduced under 13):

```markdown
* Static fields in legacy forms are now automatically sanitised to prevent XSS,
  if necessary plugins that require unsafe or malformed HTML markup in static fields
  need to be updated to use $field->set_allow_xss(true).
```

Hacking on the form to use `set_allow_xss` works (didn't care for input sanity) but I'm not sure what the advantage of using `static` over `html` is. Is it mostly to allow the field to render like a quickform element with a label?

Is this something that you'd accept a PR for? If so how do you guys usually handle breaking changes to Totara APIs do you use conditionals based on something like `isset($CFG->totara_build)`, `file_exists('/server')` or git branching etc?

Thanks for the amazing work on this plugin :)

PR #347, opened 2022-05-10
Use set_allow_xss to mark elements as safe for newer Totara versions

LiamKearn:
Not sure if you care to maintain Totara compatibility.

These were the only three cases of REXEXP: `('|")static('|")` (mform static elements) I could find in the codebase. This should resolve the issue with static mform elements not allowing html in Totara 13+: #343.

I didn't properly look at xss risks but at a quick glance at each of the changed elements:
 1. QR is entirely generated and should not contain end-user editable content.
 2. I'm not entirely sure whats the deal with xss from languages strings e.g. `setupfactor:linklabel`. Even if lang strings don't have some other form of cleaning, I'd imagine that authorisation to edit lang strings should be a `xss_risk` anyways.
 3. The User's username does get inserted into the table the last element displays.

Checking CFG doesn't feel like a really neat way to have compatibility for both if anyone has a better idea let me know.

### DIFF
--- a/factor/totp/classes/factor.php
+++ b/factor/totp/classes/factor.php
@@ -123,7 +123,13 @@ class factor extends object_factor_base {
      * {@inheritDoc}
      */
     public function setup_factor_form_definition_after_data($mform) {
-        global $OUTPUT, $PAGE, $SITE, $USER;
+        global $OUTPUT, $SITE, $USER;
+
+        /**
+         * Array of elements to allow XSS on when we're running Totara.
+         * @var array<\MoodleQuickForm_static>
+         */
+        $xssallowedelements = [];
 
         $mform->addElement('html', $OUTPUT->heading(get_string('setupfactor', 'factor_totp'), 2));
         $mform->addElement('html', \html_writer::tag('p', get_string('info', 'factor_totp')));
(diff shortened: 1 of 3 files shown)


LiamKearn:
@alexmorrisnz Sorry for the bump here.

Did you come across this in your recent PR? You mentioned you guys have it on recent Totara sites.

Was also thinking of asking Totara to include the `after_require_login` patch in core.

Any thoughts?

LiamKearn:
Following on from the CFG method of detecting Totara being weird I noticed: https://github.com/blindsidenetworks/moodle-mod_bigbluebuttonbn/pull/405/files since making this patch.

Not exactly sure how Totara thought a change like this was a good idea. IMO If I was them I'd do the XSS stuff for their own core stuff not external plugins. They still pass the pen tests this way and also don't make it harder to utilise moodle plugins.

alexmorrisnz:
> Did you come across this in your recent PR? You mentioned you guys have it on recent Totara sites.

Haven't come across this, the sites I've put it on are either T12 or have consistent cleaning off so it's not an issue there.

> Was also thinking of asking Totara to include the after_require_login patch in core.

We've submitted a feature request for the required patches to be added to core over a year ago. I recommend submitting a request, more people asking might help push this along.

I'll see if I can get someone to review this.

Peterburnett:
This all looks great. The other alternative I have touched to check totara is static method/class exists checks similar to the BBB link posted. See: https://github.com/catalyst/moodle-tool_mfa/blob/a69aff1/classes/manager.php#L487
As is though, this looks fine. As a tiny cleanup, it might be idea to refactor slightly to generate all the elements, optionally check totara once and set xss, then add all the elements, so there is only 1 block of totara compatability. Please let me know if you want to touch this up, otherwise I can make a tiny followup patch to tighten that slightly.

factor/totp/classes/factor.php
```diff
@@ -184,7 +190,14 @@ public function setup_factor_form_definition_after_data($mform) {
         ]);
 
         $html = $togglelink . $html;
-        $mform->addElement('static', 'enter', '', $html);
+        $xssallowedelements[] = $mform->addElement('static', 'enter', '', $html);
+
+        // Allow XSS on Totara.
+        if (method_exists('MoodleQuickForm_static', 'set_allow_xss')) {
```
LiamKearn, line 196: Used an array (and already created elements) out of laziness to avoid having to deal with ordering.

LiamKearn:
Just seconded that feature request to Totara.

I've pushed two new options (one per file/commit) let me know if either of those would work and I'll clean up the branch.

One (slightly more expensive) option could be to just walk all the factor forms if the method exists removing the need for these patches in the future (and disregarding the xss stuff).

Otherwise I'm very happy to defer to your patch here as you guys have a lot more experience with this stuff then me.

LiamKearn:
Also just to be clear I checked over that Totara patch and all of the code is MOODLE licensed but you guys may want to double check.

Issue #208, opened 2020-06-09
Order icons don't order the list on the Manage MFA page


golenkovm:
Order icons don't order the list.
Clicking on them does nothing from GUI end:

![image](https://user-images.githubusercontent.com/22066290/84118826-dd119000-aa76-11ea-9d3c-520b8dfa5ea2.png)

Peterburnett:
Hmm this was reported by another user as well, but I was unable to replicate. Are you able to nail down the reproduction steps? It may be due to the layout of the table itself, with gaps between active factors. Just a guess though

Peterburnett:
Also reported in #182

golenkovm:
@Peterburnett I believe it happens when you enable the factor on its own setting page (not via eye-icon). Then you can't move the factor up and down.
When the factor is enabled via eye-icon, all works fine.

Peterburnett:
Sounds like its the same issue as #130

golenkovm:
Yep

Issue #182, opened 2020-04-27
Ordering does not work as intended.


karnakitab:
I enabled Email, TOTP and No Setup factor. While trying to make TOTP Authentication at the top or default method, the up down buttons do not work. Even giving more weightage (=200) does not make any difference. Please check the issue.

Peterburnett:
Hi @karnakitab, Can you  provide any more details on this issue, such as Moodle version, MFA configuration, MFA plugin version etc., and reproduction steps? I haven't been able to reproduce this issue on master. Once I have some more details on your setup, I can dig further into this issue. 

We are also welcome to PR's if you identify the issue and are able to fix it yourself. Thanks

PR #349, opened 2022-05-17
Fix factor ordering not working when factors are enabled in their settings forms

### LINKED ISSUE #130: When disabling / enabling factors from the factor settings page the order can get messed up
ie clicking the eyeball does extra logic which is not triggered from the normal checkbox

dryj:
If you enable a factor by using the checkbox in the factor's settings form, it does not add the factor to the `tool_mfa/factor_order` setting, which then causes the factor table to not display the enabled factor at the top, and it cannot be moved around the table.

Fixes #130 #182 #208

### DIFF
--- a/factor/admin/settings.php
+++ b/factor/admin/settings.php
@@ -25,9 +25,13 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$settings->add(new admin_setting_configcheckbox('factor_admin/enabled',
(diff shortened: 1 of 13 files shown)


Peterburnett (approved):
This feels like it should be a shared function, DRY and all that, but with the amount of boilerplate that the settings pages already have, I'm not that worried about it. Thanks for fixing an annoying UX issues that we haven't had the time to fix!

PR #355, opened 2022-09-06
Remove blocks from auth page #352

### LINKED ISSUE #352: Site-wide blocks appear on TOTP & Email factor page
I have found that site-wide blocks appear on the TOTP and email factor pages, and because the user is formally authenticated with Moodle, this can leak sensitive information.

I'm not hugely familiar with Moodle core, but having loooked at how the quiz module does it, I think the fix would be to follow the set_pagelayout() call with a call to show_only_fake_blocks()

```
$PAGE->set_pagelayout('secure');
$PAGE->blocks->show_only_fake_blocks();
```

These two places would need changing:
https://github.com/catalyst/moodle-tool_mfa/blob/master/auth.php#L36
https://github.com/catalyst/moodle-tool_mfa/blob/master/factor/email/email.php#L38

Search for the issue:
https://github.com/catalyst/moodle-tool_mfa/search?q=set_pagelayout+secure

brendanheywood:
Resolves #352

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -34,6 +34,7 @@ $context = context_user::instance($USER->id);
 $PAGE->set_context($context);
 $PAGE->set_url('/admin/tool/mfa/auth.php');
 $PAGE->set_pagelayout('secure');
+$PAGE->blocks->show_only_fake_blocks();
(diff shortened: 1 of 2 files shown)


Peterburnett:
Only issues are phpdoc. Should be cleaned up at some point, but not a blocker now

PR #357, opened 2022-09-07
356 spring clean

keevan:
phpcs cleanup and phpdoc fixes across the plugin

Skips:
- test case coverage as that will take more time and should be correct.
- test namespace changes
- privacy warnings for `_get_reason`


Peterburnett (approved):
All CI is passing, good to go on

PR #361, opened 2022-10-06
fix(totp): a unique factor will only be stored at most once

### LINKED ISSUE #360: Prevent TOTP factor record for the same key to be stored more than once
Report of it potentially storing more than once, which could happen as part of a double submission.

keevan:
Resolves #360

### DIFF
--- a/factor/totp/classes/factor.php
+++ b/factor/totp/classes/factor.php
@@ -338,7 +338,6 @@ class factor extends object_factor_base {
      *
      * @param stdClass $data
      * @return stdClass the factor record, or null.
-     * @throws \moodle_exception if the TOTP factor already exists in the DB
      */
     public function setup_user_factor($data) {
         global $DB, $USER;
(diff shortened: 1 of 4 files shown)


factor/totp/classes/factor.php
```diff
@@ -354,6 +355,16 @@ public function setup_user_factor($data) {
...
             $row->revoked = 0;
 
+            // Check if a record with this secret already exists for the same user. If so, throw an exception.
+            $exists = $DB->record_exists('tool_mfa', [
+                'userid' => $row->userid,
+                'secret' => $row->secret,
+                'factor' => $row->factor,
+            ]);
+            if ($exists) {
+                throw new \moodle_exception('error:alreadyregistered', 'factor_totp');
```
Peterburnett, line 365: Could this just be an alert notification and instead of an exception? Little less disruptive to the user

keevan: Yes, I'll change this.

keevan: So follow up question. Should `setup_user_factor` return the existing record? Or `null` in this case?

If it returns null, then it throws a different error `Can not setup factor` and we end up in the same place, with a bit more noise.

So I've opted to return the record instead, and set the notification to a warning so the user is redirected to the list of factors instead of back to the TOTP setup page.

Please let me know what you expect them to land on if it already exists.

factor/totp/classes/factor.php
```diff
@@ -354,6 +354,17 @@ public function setup_user_factor($data) {
             $row->lastverified = 0;
             $row->revoked = 0;
 
+            // Check if a record with this secret already exists for the same user. If so, throw an exception.
```
Peterburnett, line 357: Comment outdated

factor/totp/tests/factor_test.php
```diff
@@ -113,4 +117,27 @@ public function test_validate_code() {
...
+        set_config('enabled', 1, 'factor_totp');
+        $totpfactor = \tool_mfa\plugininfo\factor::get_factor('totp');
+        $totpdata = [
+            'secret' => 'fakekey',
+            'devicename' => 'fakedevice',
+        ];
+        $record = $totpfactor->setup_user_factor((object) $totpdata);
+
+        // Trying to add the same TOTP should return the existing record.
+        $anotherecord = $totpfactor->setup_user_factor((object) $totpdata);
```
Peterburnett, line 140: Here there should also be a DB->count_records = 1 perhaps?

factor/totp/classes/factor.php
keevan: Done

factor/totp/tests/factor_test.php
keevan: Done

Peterburnett (approved):

PR #373, opened 2023-02-26
Totara 17.3+ patchless support.

LiamKearn:
## Changes
This patch attempts to watch the newly introduced Totara hooks:
 - `\core\hook\after_require_login`
 - `\core\hook\after_config`

It will forward any calls to them along to the existing lib.php callbacks (`tool_mfa_after_config` & `tool_mfa_after_require_login`) that check for authentication.

These hooks were introduced as part of TL-35168. If you are not familar with Totara hooks documentation is available here: https://help.totaralearning.com/display/DEV/Hooks+developer+documentation

This will authenticate users on Totara (17.3+) without the need for core patches.

All changes in this patch are gated with a class_exists check for `\totara_core\hook\base` to ensure MOODLE compatibility.

---
## Three things to consider

1.	It seems the plan in the future is for these hooks to automatically search for functions in lib.php as normal to preserve functionality with MOODLE plugins. That future plan will hopefully render this change obsolete and make this plugin work with Totara without any changes.
	
	The source for that plan is a code comment:
	`TODO - PLATFORM-117 to create a watcher to enable those functions via this hook` in one of the 
	newly introduced hooks.

1. I'm sure the `\class_exists()` in hooks.php is not needed but it can't hurt to be explicit.

1.	With this patch lib.php callbacks could potentially be called before the plugin is installed or enabled. This is different to the MOODLE hook style with `get_plugins_with_function` which does something like:
	```php
	if (!isset($installedplugins[$plugin])) {
	    // Plugin code is still present on disk but it is not installed.
	    unset($pluginfunctions[$plugintype][$plugin]);
	    continue;
	}
	```
	
	At the moment the `tool_mfa\manager::require_auth` invoked via the lib.php callbacks handles this in it's call to: `\tool_mfa\manager::is_ready`. However if anything new was introduced directly to lib.php callbacks that requires installed plugin functionality it may create issues.
	
	I think it's safe to ignore for now especially since Totara seemingly plans to make this obsolete in the future.

### DIFF
--- a/classes/watcher/login_watcher.php
+++ b/classes/watcher/login_watcher.php
@@ -1,0 +1,67 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_mfa\watcher;
+
+defined('MOODLE_INTERNAL') || die();
+
+use core\hook\after_require_login;
+use core\hook\after_config;
+use totara_core\hook\base;
+
+// Hooks are Totara specific.
+if (!class_exists(base::class)) {
+    return;
+}
+
+require_once($CFG->dirroot . '/admin/tool/mfa/lib.php');
+
+/**
+ * Watches Totara hooks to ensure a user is authenticated.
+ *
+ * Forwards on the events until Totara implements:
+ * `TODO - PLATFORM-117 to create a watcher to enable those functions via this hook`
+ *
+ * @package     tool_mfa
+ * @author      Liam Kearney <liam@sproutlabs.com.au>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class login_watcher {
+    /**
+     * Forwards calls to tool_mfa_after_require_login.
+     *
+     * @param after_require_login $hook
+     */
+    public static function ensure_authenticated(after_require_login $hook) {
+        tool_mfa_after_require_login(
+            $hook->courseorid,
+            $hook->autologinguest,
+            $hook->cm,
+            $hook->setwantsurltome,
+            $hook->preventredirect
+        );
+    }
+
+    /**
+     * Forwards calls to tool_mfa_after_config.
+     *
+     * @param after_config $hook
+     */
+    public static function check_authenticated(after_config $hook) {
+        tool_mfa_after_config();
+    }
+}
--- a/db/hooks.php
+++ b/db/hooks.php
@@ -1,0 +1,52 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Definition of MFA sub-plugins (factors).
+ *
+ * @package     tool_mfa
+ * @author      Liam Kearney <liam@sproutlabs.com.au>
+ * @copyright   Catalyst IT
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+defined('MOODLE_INTERNAL') || die();
+
+use core\hook\after_require_login;
+use core\hook\after_config;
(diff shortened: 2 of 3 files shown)


LiamKearn:
I'm defeated by CI's required parameter documentation in MOODLE 4+. I don't develop enough MOODLE/Totara at the moment to justify setting codesniffs up locally in VIM as I had it in vscode. I'm not entirely sure why it fails with (no function documentation) and only after I add it does it complain about missing parameters (especially when those parameters are typed). IMO CI checks should be declarative not imperative.

Might come back to this another day but hoping I can be a cheeky boy and leave it to you to fix as maintainers if want to include this 😉

LiamKearn:
@danmarsden is there any chance you/someone could take a look at this in the next few weeks?

I saw you created an issue around T17 in this plugin recently.

At this point the CI linting failures seem unrelated to files touched in this patch:
Screenshot 2023-03-10 at 3 24 12 pm

Peterburnett:
@LiamKearn Thanks for the contributions, and the very detailed description. Ill try and squeeze this onto my plate in the next week or two to get it polished off and merged. Cant say Im overly thrilled with the direction totara is taking, but happy to match to it.

classes/watcher/login_watcher.php
```diff
@@ -0,0 +1,63 @@
...
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_mfa\watcher;
+
+defined('MOODLE_INTERNAL') || die();
+
+use core\hook\after_require_login as after_require_login_hook;
+use core\hook\after_config as after_config_hook;
+
+if (class_exists('\totara_core\hook\base')) {
```
keevan, line 24: Could we flip the bit here and make it return early instead? The benefit in doing so is so the class and any extra code blocks are indented at the same levels as other files, and if conditions change or is removed in the future, it won't cause a big diff that is effectively only indentation changes.

```
if (!class_exists('\totara_core\hook\base')) {
     return;
}
```

The same could similarly apply for hooks.php which would keep $watchers at the edge instead of indented due to the surrounding conditional block.

classes/watcher/login_watcher.php
```diff
@@ -0,0 +1,63 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace tool_mfa\watcher;
+
+defined('MOODLE_INTERNAL') || die();
+
+use core\hook\after_require_login as after_require_login_hook;
+use core\hook\after_config as after_config_hook;
```
keevan, line 22: It would be better to `use` these classes without aliasing them.

As per the Moodle coding standards here, this is discouraged and should only be used when resolving name conflicts: https://moodledev.io/general/development/policies/codingstyle#further-notes

> Do not alias class imports (use XXX as YYY;) unless it is absolutely required to resolve a conflict.

db/hooks.php
```diff
@@ -0,0 +1,45 @@
...
+ * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+defined('MOODLE_INTERNAL') || die();
+
+// Hooks are Totara specific.
+if (class_exists('\totara_core\hook\base')) {
+    $watchers = [
+        [
+            'hookname' => '\core\hook\after_require_login',
```
keevan, line 32:
```suggestion
            'hookname' => core\hook\after_require_login::class,
```
Would this also work? The benefit here is some IDEs such as VSCode and PHPStorm pick up on these symbols, and if they resolve to a real class, it will display the relevant tooltip / jump-to / definition, which could be an extra check to ensure a real class is being referenced. Otherwise, if it exists as a string, it might not be as useful. 

That said, this is Totara specific and I am not well versed in these new hooks declarations, and the way you have it might be more consistent with how they've set up their examples.

It could similarly apply for after_config and the `totara_core\hook\base::class` check.

What are your thoughts here?

classes/watcher/login_watcher.php
LiamKearn: Done for both touched files, I'm not sure if short-circuiting has any effect be it autoloaded files or included ones? Let me know if so and I'll revert `db/hooks.php`

classes/watcher/login_watcher.php
LiamKearn: Done, Aliasing is generally bad in all languages, frameworks I've worked with but after_config is a terrible classname (regardless of the context provided by it's namespace).

db/hooks.php
LiamKearn: Good suggestion, absolutely all for magic classnames!

I've converted the callable syntax to use `array<class-string, string>` instead as well.

LiamKearn:
@Peterburnett 
> @LiamKearn Thanks for the contributions, and the very detailed description. Ill try and squeeze this onto my plate in the next week or two to get it polished off and merged. Cant say Im overly thrilled with the direction totara is taking, but happy to match to it.

Yes completely agree, I'm not a fan of hooks, Totara feels like a [wheel design shop](https://docs.moodle.org/401/en/Event_monitoring) in some regards.

If you're commenting on the fact that we have to manually watch these hooks as of now that should change in the future. I think they're doing it as a two step plan:
 1. Add the hooks <- We're here now
 1. Call the hooks to preserve moodle compat without developers having to watch them

---
 
@keevan
Thank you for taking the time to review, I've improved everything a bit based on your feedback, let me know if there is anything further.

classes/watcher/login_watcher.php
keevan: > after_config is a terrible classname

Totally

LiamKearn:
I intentionally added [that typo](https://github.com/catalyst/moodle-tool_mfa/compare/c3edf7affbc30a8cd0fbeb48c71c031629de593e..a4853de4e8f8fae906789c9568019593e6b08c76) to test something earlier didn't realise it was commited, glad it was picked up before a merge :)

LiamKearn:
Hi All

Apologies for bump, any chance someone can review/trigger CI again (I think failures are unrelated).

It would be really nice to no longer have to patch for such a standard security measure.

danmarsden:
Thanks Liam, now that Moodle is bringing this into their core release we've created a Totara specific branch for Totara support - If you got a chance to rebase and fix the conflict it would be great, otherwise we'll try to get to it ourselves sometime.

PR #374, opened 2023-03-09
Portuguese version

Lucas040201:
Hello, I'm using this plugin and I added a Brazilian Portuguese translation. I would like you to make available to other users if possible


danmarsden:
Hi there! Thanks for looking into this, our plugins all use moodles official translation site (amos) and do not store translations inside the codebase. More information on this process is here: https://docs.moodle.org/401/en/Translating_plugins

It would be great if you could submit the changes there!

PR #375, opened 2023-03-16
Properly check source auth page when detecting redir loops

Peterburnett:


classes/manager.php
```diff
@@ -537,9 +533,26 @@ public static function should_require_mfa($url, $preventredirect) {
...
+            return self::NO_REDIRECT;
+        }
+
         return self::REDIRECT;
     }
 
+    /**
+     * Clears the redirect counter for infinite redirect loops. Called from auth.php when a valid load is resolved.
+     *
+     * @return void
```
keevan, line 547: No need to include this if it doesn't return anything.

classes/manager.php
```diff
@@ -537,9 +536,25 @@ public static function should_require_mfa($url, $preventredirect) {
         // Set referer after checks.
         $SESSION->mfa_redir_referer = get_local_referer(true);
 
+        // Dont redirect if already on auth.php.
```
keevan, line 539: Dont -> Don't

keevan (approved):

Issue #330, opened 2022-02-28
No-setup/-other Factor makes Authenticator app settings disapear


agru:
I've setup the MFA tool (in Moodle 3.9.12+) and want to offer my users the possibility to use TOTP via the Authenticator app factor. Enabling only this factor disables login at all, as the users do not have TOTP setup until now. For already logged in users (logged in before the app factor is enabled) there is an option in their personal settings to setup TOTP. But this was just a test - in reality the users are not logged in and are not able to login any more. So I've added the No-setup / No other factor, as this should solve the problem according to the documentation. Fine, they are able to login although the Auth-app-factor is also enabled, but enabling this second factor makes the link in the personal settings of the users disappear. Where can they now configure their TOTP-settings?

zozo1977:
Same here. Trying to figure this out for hours. I already thought I'm missing something in the plugin setup because I didn't notice this issue message at first.

AnnieHy:
same as above written:
I would like that users (e.g. students) are able to decide for themselves if and which authentication they use. Admins have to use the MFA. The settings for admins are fine, but how can I setup mfa so user can choose by themselves? According to https://github.com/catalyst/moodle-tool_mfa#no-setup-factor it is recommended to use the no-setup factor, but this factor does not appear in the overview. 

Since the plugin is quite complex, my question would be if there are more examples of how to use it with different  scenarios. The explanations on https://github.com/catalyst/moodle-tool_mfa are not 100% explaining for me. Thank you for your advice!

Peterburnett:
Hi all,

This issue has just come to my attention. The reason the admin menu is disappearing is that MFA is reading that the user currently has a factor that can ONLY pass for them, and so setting up another factor is redundant, and the menu won't show. This is to combat scenarios where MFA should ONLY be rolled out some some system roles for example, while other roles are completely transparent. There may be an issue with this behaviour in conjunction with the no-setup factor, but I don't have time to get into it right now.

An investigation into code or a community PR for this would be greatly appreciated in the interim.

Thanks!

90084:
Some minor comments from my research.
Preferences used to show in version 2020021401, but they're not showing on latest version 2023031600. I noticed some changes in manager.php:possible_factor_setup() back in 2020-03-20, I believe that introduced the change.

Suggested workarounds:
- Provide a direct link to the MFA preferences with https://yourdomain.tld/admin/tool/mfa/user_preferences.php (yes, the user does have access to this page)
[...]

PR #381, opened 2023-05-24
Skip factor nosetup while checking pass-factors in get_enabled_factors()

agru:
When factor totp and factor nosetup are active there is no link to setup MFA in the user settings. This is due to nosetup being a factor with STATE_PASS and a necessary weight of 100. As factor nosetup is a pseudo factor it can not be counted towards the question whether the user can pass with this factor.

This solves issue #330


agru:
Will be added to a further PR

PR #382, opened 2023-05-24
Skip factor nosetup while checking pass-factors in get_enabled_factors() & Increase key length to 128 & German TOTP language file

agru:
When factor totp and factor nosetup are active there is no link to setup MFA in the user settings. This is due to nosetup being a factor with STATE_PASS and a necessary weight of 100. As factor nosetup is a pseudo factor it can not be counted towards the question whether the user can pass with this factor. This solves issue https://github.com/catalyst/moodle-tool_mfa/issues/330

The secret key length of the TOTP factor should be 128 bits as requested by RFC 4226, section 4, requirement 6. As at least one TOTP app, FreeOTP, shows nasty warnings about a weak key when using only 80 bit key length. This commit increases the key length to 128 bit without breaking the compatibility to already existing 80 bit keys in the database for users that have setup TOTP before this change. This solves issue https://github.com/catalyst/moodle-tool_mfa/issues/383

Additionally included is a German translation file for the TOTP factor

### DIFF
--- a/classes/manager.php
+++ b/classes/manager.php
@@ -792,6 +792,7 @@ class manager {
         // Check if there are enough factors that a user can ONLY pass, if so, don't display the menu.
         $weight = 0;
         foreach ($factors as $factor) {
+            if ( $factor->name == "nosetup") { continue; }
             $states = $factor->possible_states($USER);
             if (count($states) == 1 && reset($states) == \tool_mfa\plugininfo\factor::STATE_PASS) {
                 $weight += $factor->get_weight();
--- a/factor/totp/classes/factor.php
+++ b/factor/totp/classes/factor.php
@@ -324,13 +324,14 @@ class factor extends object_factor_base {
     }
 
     /**
-     * Generates cryptographically secure pseudo-random 16-digit secret code.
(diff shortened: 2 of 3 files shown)


danmarsden:
Thanks @agru - can you please remove the translation file from the PR - best place for that is in amos rather than maintained within the plugin lang directory.

classes/manager.php
```diff
@@ -792,6 +792,7 @@ public static function possible_factor_setup() {
         // Check if there are enough factors that a user can ONLY pass, if so, don't display the menu.
         $weight = 0;
         foreach ($factors as $factor) {
+            if ( $factor->name == "nosetup") { continue; }
```
Peterburnett, line 795: Im not sure this is the best way to approach this, it feels the possible_states in the factor for this. Eg it should return a STATE_NEUTRAL always instead of a STATE_PASS, with a decent explainer comment that to support optional rollouts correctly, the state needs to be exposed as NEUTRAL otherwise the menu will not display.

danmarsden:
I've pushed through a different method for addressing the nosetup issue in #417 - closing this one off, but feel free to submit another PR for the key size in the totp factor.

PR #388, opened 2023-06-01
Fix string confusion after factor creation

Peterburnett:


classes/local/factor/object_factor_base.php
```diff
@@ -281,7 +281,17 @@ public function revoke_user_factor($factorid = null) {
...
         if (!empty($factorid)) {
+            // If we have an explicit factor id, this means we need to be careful about the user.
             $params = ['id' => $factorid];
+            $existing = $DB->get_record('tool_mfa', $params);
+            if (empty($existing)) {
+                return false;
+            }
+            $matchinguser = $existing->userid == $USER->id;
+            if (!(is_siteadmin() || $matchinguser)) {
+                // We aren't admin, and this isn't our factor.
```
keevan, line 292: Perhaps it might be easier to skim through if the condition was
```
if (!is_siteadmin() && !$matchinguser) {
```
 (e.g. less OR and !() conditions)

dmitriim: +1 for this one. Took me few seconds to process` !(is_siteadmin() || $matchinguser)`

Peterburnett: Im in the opposite camp, purely because "not siteadmin or user" sounds more natural in my brain than "not siteadmin and not user". Im happy to change it, but I find the current easier to grok.

Peterburnett: OK ill change it

user_preferences.php
```diff
@@ -54,9 +54,12 @@
 if (!empty($action)) {
     if ($factorid != 0) {
         $instance = \tool_mfa\plugininfo\factor::get_instance_from_id($factorid);
-        $factor = \tool_mfa\plugininfo\factor::get_factor($instance->factor);
-        $string = $factor->get_display_name().' - '.$instance->label;
-        echo $OUTPUT->notification(get_string('factor'.$action, 'tool_mfa', $string), 'notifysuccess');
+        // Confirm factor is valid for the accessing user.
```
dmitriim, line 57: I'm reading this file and can't understand why we need $userid. Should we just remove all references?

classes/local/factor/object_factor_base.php
```diff
@@ -281,7 +281,17 @@ public function revoke_user_factor($factorid = null) {
...
 
         if (!empty($factorid)) {
+            // If we have an explicit factor id, this means we need to be careful about the user.
             $params = ['id' => $factorid];
+            $existing = $DB->get_record('tool_mfa', $params);
+            if (empty($existing)) {
+                return false;
+            }
+            $matchinguser = $existing->userid == $USER->id;
+            if (!is_siteadmin() && !$matchinguser) {
```
dmitriim, line 291: is there any chance we can add tests to cover new changes?

user_preferences.php
Peterburnett: Yeah this is a hang over from the very early implementation. Im not sure what the original intent was besides excluding guests, but I agree that it seems unnecessary. Ill kill it off

classes/local/factor/object_factor_base.php
Peterburnett: There currently isnt coverage of the abstract class implementation, which is why I haven't already, as its a larger piece to embark on, but I will get that ball rolling now

tests/object_factor_base_test.php
```diff
@@ -0,0 +1,75 @@
...
+        $factorinstance = $totpfactor->setup_user_factor((object) $totpdata);
+        $totpdata2 = [
+            'secret' => 'fakekey2',
+            'devicename' => 'fakedevice2',
+        ];
+        $totpfactor->setup_user_factor((object) $totpdata2);
+
+        $this->assertFalse((bool) $factorinstance->revoked);
+        $this->assertEquals(2, count($totpfactor->get_active_user_factors($user)));
+
```
dmitriim, line 50: double empty line here. CI should be mad now :)

tests/object_factor_base_test.php
```diff
@@ -0,0 +1,74 @@
...
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Tests for base factor implementation methods.
```
dmitriim, line 18: I think you don't  need comment on file anymore. This comment should be moved to a class itself.

dmitriim: @Peterburnett I can't see file doc block moved to a class doc block. Have you forgotten to push it?

PR #400, opened 2023-06-09
Removetotaracode

danmarsden:
dropping in a quick patch to remove Totara specific code from the 4.0 branch - we support Totara on the older 3.5 branch now.
@alexmorrisnz do you want to peer review this and merge it if it looks ok?


factor/totp/classes/factor.php
```diff
@@ -156,13 +153,11 @@ public function setup_factor_form_definition_after_data($mform) {
         $qrcode = $this->generate_qrcode($secret);
 
         $html = \html_writer::tag('p', $qrcode);
-        $xssallowedelements[] = $mform->addElement('static', 'scan', get_string('setupfactor:scan', 'factor_totp'), $html);
```
alexmorrisnz, line 159: Needs to be changed to only remove the `$xssallowedelements[] = ` section, otherwise we remove the form elements we want. Same on line 165 and 200

danmarsden: grr - not sure what I was doing there - fixed that up.

PR #408, opened 2023-06-15
Remove tags from devicename

TomoTsuyuki:
Currently, we can add tags in the device name for TOTP.
This patch prevents adding the tags for device name.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -132,5 +132,15 @@ function xmldb_tool_mfa_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2021021900, 'tool', 'mfa');
     }
 
+    if ($oldversion < 2023031601) {
+        // Update the label field for TOTP to strip tags.
+        $records = $DB->get_records_select('tool_mfa', $DB->sql_like('label', '?'), ['%<%'], '', 'id, label');
+        foreach ($records as $record) {
+            $DB->set_field('tool_mfa', 'label', strip_tags($record->label), ['id' => $record->id]);
+        }
+        // Mfa savepoint reached.
+        upgrade_plugin_savepoint(true, 2023031601, 'tool', 'mfa');
+    }
+
     return true;
 }
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -136,7 +136,7 @@ function xmldb_tool_mfa_upgrade($oldversion) {
         // Update the label field for TOTP to strip tags.
         $records = $DB->get_records_select('tool_mfa', $DB->sql_like('label', '?'), ['%<%'], '', 'id, label');
         foreach ($records as $record) {
-            $DB->set_field('tool_mfa', 'label', strip_tags($record->label), ['id' => $record->id]);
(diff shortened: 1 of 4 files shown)


factor/totp/classes/factor.php
```diff
@@ -347,7 +347,7 @@ public function setup_user_factor($data) {
             $row->userid = $USER->id;
             $row->factor = $this->name;
             $row->secret = $data->secret;
-            $row->label = $data->devicename;
+            $row->label = strip_tags($data->devicename);
```
dmitriim: I'd also recommend to check if we can clean data before we save it to DB.

dmitriim: ah we are saving data later... 
any reason for not using clean_param function?

dmitriim: What about already existing infected records? Do we need an upgrade script? 
Should we apply something before rendering on a screen? 
Is there any other methods potentially could be compromised the same way?

TomoTsuyuki: The posted record is coming to `$data` to this function, so I think it's better to use just `strip_tags` here.
I can update the code if you have better reason.

TomoTsuyuki: I added update script.
I tested other user inputs (`Mobile phone SMS` and `Security Key`) and they looked fine.

db/upgrade.php
```diff
@@ -132,5 +132,15 @@ function xmldb_tool_mfa_upgrade($oldversion) {
         upgrade_plugin_savepoint(true, 2021021900, 'tool', 'mfa');
     }
 
+    if ($oldversion < 2023031601) {
+        // Update the label field for TOTP to strip tags.
+        $records = $DB->get_records_select('tool_mfa', $DB->sql_like('label', '?'), ['%<%'], '', 'id, label');
+        foreach ($records as $record) {
```
dmitriim, line 138: This loop could take a long time if we have a lot of users. Anything we could do to speed up? or maybe move to a background?

factor/totp/classes/factor.php
dmitriim: Is it possible to add tests to cover that?

db/upgrade.php
TomoTsuyuki: I know it can be longer but this is a setting update for the users who use html tag for devicename (actually just checking `<` though).
I don't think many users are coming to this record update.
I think it's overkill to use background.

factor/totp/classes/factor.php
TomoTsuyuki: It looks the `action.php` for the `TOTP` didn't clean tags when posting, but clean up for the others.
I'm not sure how I can make test for `action.php` as it's page, not internal function.

dmitriim: as we cleaning data on save we should test that. So in out test: we feed  setup_user_factor devicename with HTML and script tags  and then verify that data was sanitised after creating/updating.

TomoTsuyuki: Test has been added.

PR #409, opened 2023-06-15
Remove tags from devicename

TomoTsuyuki:
Currently, we can add tags in the device name for TOTP.
This patch prevents adding the tags for device name.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -136,7 +136,7 @@ function xmldb_tool_mfa_upgrade($oldversion) {
         // Update the label field for TOTP to strip tags.
         $records = $DB->get_records_select('tool_mfa', $DB->sql_like('label', '?'), ['%<%'], '', 'id, label');
         foreach ($records as $record) {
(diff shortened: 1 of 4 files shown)


golenkovm:
Hi @TomoTsuyuki 

Can we please make sure the patch is robust against XSS attacks? As I can see there are some caveats mentioned here  https://www.php.net/manual/en/function.strip-tags.php 

Cheers,
Misha

TomoTsuyuki:
> Hi @TomoTsuyuki
> 
> Can we please make sure the patch is robust against XSS attacks? As I can see there are some caveats mentioned here https://www.php.net/manual/en/function.strip-tags.php
> 
> Cheers, Misha

Updated by using `htmlspecialchars` function.

danmarsden:
is there a reason the standard Moodle Api clean_param() isn't being used? - all good if there is :-)

TomoTsuyuki:
> is there a reason the standard Moodle Api clean_param() isn't being used? - all good if there is :-)

Thanks for your suggestion.
I updated the code by `clean_param()`.

PR #417, opened 2023-07-06
Check nosetup factor correctly so that menu displays.

### LINKED ISSUE #330: No-setup/-other Factor makes Authenticator app settings disapear

### LINKED ISSUE #391: Preferences link to setup MFA does not appear
Hello,

I would like users to be able to setup TOTP if they want. If they do not want, they should be able to login normally.

My problem is:
- For admins: the option "Multi-factor authentication preferences" does not appear in user preferences if there is a way to log in without needing to set up TOTP.
- For non admins: The option never appears. But they are able to access directly "/admin/tool/mfa/user_preferences.php".

How can I make so that that option appears always for all users, so they are able to setup TOTP if they want?

danmarsden:
Fix #330 Fix #391

### DIFF
--- a/factor/nosetup/classes/factor.php
+++ b/factor/nosetup/classes/factor.php
@@ -92,15 +92,8 @@ class factor extends object_factor_base {
      * @return void
      */
     public function possible_states($user) {
-        // Check if user has any other input or setup factors active.
-        $factors = \tool_mfa\plugininfo\factor::get_active_other_user_factor_types($user);
-        foreach ($factors as $factor) {
-            if ($factor->has_input() || $factor->has_setup()) {
-                return [\tool_mfa\plugininfo\factor::STATE_NEUTRAL];
-            }
-        }
-
-        return [\tool_mfa\plugininfo\factor::STATE_PASS];
+        // We return Neutral here because to support optional rollouts it needs to report neutral or the menu to setup will not display.
+        return [\tool_mfa\plugininfo\factor::STATE_NEUTRAL];
     }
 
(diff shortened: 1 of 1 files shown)


danmarsden:
@Peterburnett can you please take a look at this one? - I think this is the right way to fix #330 #391 instead of the method used in #382

Peterburnett:
So my only concern here is this removes the use case where you want only a certain cohort of users to be able to use MFA. Eg administrators etc. The intent of the splitting is that you don't see the menu if you would be awarded points and able to authenticate. I agree that it is causing more headaches for some users than it might be worth...

Peterburnett:
Hmm looks like I already thought about this a while ago and mentioned it in #382. I think this is good to go. Benefit outweighs the drawbacks.

Peterburnett (approved):

danmarsden:
thanks - backported to 35_STABLE branch too.

Issue #383, opened 2023-05-25
TOTP key length should be increased from 80 to at least 128 bit


agru:
The TOTP factor returns a string of 16 characters. This corresponds, after Base32 decoding, to an 80-bit secret key. However, according to the requirement for the OTP algorithm in RFC4226 (see Section 4, R6 there), a 128-bit key is a MUST. Most TOTP apps accept the 80-bit key, but the app FreeOTP as an example does not. Here our users get the following message after scanning the QR code: "Token is unsafe! The token you are attempting to add contains weak cryptographic parameters. Use of this token is stronly discouraged! Please alert your token provider". The key length of the TOTP factor should be increased to 128 bits to avoid this nasty warning.

agru:
PR #382 solves this

toheine:
Great! Thank you. I fell over this exact error this morning while setting up 2FA in Moodle with FreeOTP, so I'm super happy about this solution!

danmarsden:
Thanks team - I'm closing this as "won't fix" in the older branches for now - Moodle now includes the mfa plugin in the core release and so feature/improvements need to land in Moodle core "first". We can backport stuff that lands upstream in Moodle core - so if you get this change into Moodle's core release, feel free to re-open this again for further consideration.

PR #424, opened 2023-08-08
Increase TOTP key length to 128

agru:
The secret key length of the TOTP factor should be 128 bits as requested by RFC 4226, section 4, requirement 6. As at least one TOTP app, FreeOTP, shows nasty warnings about a weak key when using only 80 bit key length. This commit increases the key length to 128 bit without breaking the compatibility to already existing 80 bit keys in the database for users that have setup TOTP before this change. This solves issue #383

### DIFF
--- a/factor/totp/classes/factor.php
+++ b/factor/totp/classes/factor.php
@@ -314,13 +314,14 @@ class factor extends object_factor_base {
     }
 
     /**
-     * Generates cryptographically secure pseudo-random 16-digit secret code.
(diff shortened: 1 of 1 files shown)


danmarsden:
as mentioned on #383  closing this as "won't fix" in the older branches for now - Moodle now includes the mfa plugin in the core release and so feature/improvements need to land in Moodle core "first". We can backport stuff that lands upstream in Moodle core - so if you get this change into Moodle's core release, feel free to re-open this again for further consideration.

PR #427, opened 2023-08-21
feat: improve email ip geo location handling

matthewhilton:
It turns out the IP location lookup can fail for a number of reasons, which stops the sending of the email code, essentially locking the user out.

This change makes handling of the location a bit better. So now if:

- it throws an exception
- returns nothing
- returns an error

It will still send the email, but the email will simply say the location was unavailable.

I also added unit tests for this, since there were no existing unit tests for the email factor.

Also, another commit was added that fixes the PHP copy paste detector error in the CI. This was purely moving the same duplicated code in 7 or so factors into one spot.


matthewhilton:
<strike>Note ci is saying failing, but this seems to be because of unrelated stuff to this change</strike>

Have added another commit to fix the CI, just a tiny bit of refactoring

Peterburnett:
This all looks good, there is just 1 more touch point in the main plugin renderer that should be cut over to the error resilient function as well

matthewhilton:
Have updated the user preferences renderer as well:

Screenshot 2023-08-22 at 13-41-32 UNKN Multi-factor authentication preferences
Screenshot 2023-08-22 at 13-41-01 UNKN Multi-factor authentication preferences

Even though the logic is kind of similar I didn't reuse the same function since the wording is different in email vs the user preferences page, so its just simpler to keep them seperate

PR #432, opened 2023-08-31
Update unit test functions

owenherbert:
- Update unit test function names

### DIFF
--- a/factor/email/tests/factor_test.php
+++ b/factor/email/tests/factor_test.php
@@ -100,12 +100,12 @@ class factor_test extends \advanced_testcase {
         $unknownstr = get_string('email:geoinfo:unknown', 'factor_email');
 
         if ($expectedunknown) {
-            $this->assertContains($unknownstr, $email);
+            $this->assertStringContainsString($unknownstr, $email);
         } else {
             // Ideally we would test the email contains the correct geo info,
             // but since the geo location of a real IP address can change
             // we instead test that the unknown message is not there.
-            $this->assertNotContains($unknownstr, $email);
+            $this->assertStringNotContainsString($unknownstr, $email);
         }
     }
 }


factor/email/tests/factor_test.php
```diff
@@ -100,12 +100,12 @@ public function test_generate_email_ip_address_location(string $ip, bool $expect
         $unknownstr = get_string('email:geoinfo:unknown', 'factor_email');
 
         if ($expectedunknown) {
-            $this->assertContains($unknownstr, $email);
+            $this->assertStringContainsString($unknownstr, $email);
```
matthewhilton, line 103: We'll want to check if this function exists first before using it. IIRC I purposely did not use assertStringContainsString because it caused the CI to fail because it didn't exist in the CI's phpunit version.

CI is still running atm but im 95% sure it'll fail because of this

factor/email/tests/factor_test.php
```diff
@@ -100,12 +100,12 @@ public function test_generate_email_ip_address_location(string $ip, bool $expect
...
         } else {
             // Ideally we would test the email contains the correct geo info,
             // but since the geo location of a real IP address can change
             // we instead test that the unknown message is not there.
-            $this->assertNotContains($unknownstr, $email);
+            $this->assertStringNotContainsString($unknownstr, $email);
```
matthewhilton, line 108: Same as above, we need to check this exists first

factor/email/tests/factor_test.php
matthewhilton: Or another option to avoid having an ugly if else check if to use strpos e.g $this->assertNotFalse(strpos()...) 

(note you can't use str_contains since that is only php8+)

matthewhilton:
Ah I spoke too soon - ci says its all good

PR #440, opened 2023-10-16
Allow for seperate admin configuration for trust token

Peterburnett:


factor/token/tests/factor_test.php
```diff
@@ -69,6 +72,24 @@ public function test_calculate_expiry_time_in_general() {
...
+        $this->setAdminUser();
+
+        // Confirm that the expiry is based on the admin duration.
+        // Duration for regular users is covered by the general calc test.
+        $method = new \ReflectionMethod($this->factor, 'calculate_expiry_time');
+        $method->setAccessible(true);
+
+        $expiry = $method->invoke($this->factor, $timestamp);
+        $this->assertGreaterThan(7 * HOURSECS, $expiry[1]);
+        $this->assertLessThan(9 * HOURSECS, $expiry[1]);
```
keevan: Are we able to deconstruct this to something a bit more readable? I'm not certain what value `$expiry[1]` refers to.

factor/token/tests/factor_test.php
```diff
@@ -69,6 +72,24 @@ public function test_calculate_expiry_time_in_general() {
         $this->assertLessThan($resetdelta + DAYSECS + 30, $expiry[1]);
     }
 
+    public function test_calculate_admin_expiry_time() {
```
keevan, line 79: Potentially phpdoc checks might complain here.

factor/token/tests/factor_test.php
Peterburnett: Updated. The expiry calculator returns a timestamp and the duration, as we need timestamp for the cookie, and one for the secret manager.=

factor/token/tests/factor_test.php
Peterburnett: fixed

keevan (approved):

PR #445, opened 2023-10-23
Do not display trust prompt on 0 weight factors

Peterburnett:


factor/token/classes/factor.php
```diff
@@ -144,14 +144,19 @@ public function global_definition_after_data($mform) {
...
 
-        if ($noproperty && $nostate) {
+        if ($noproperty && $nostate && $weight !== 0) {
             $expiryconfig = is_siteadmin() ? 'adminexpiry' : 'expiry';
             $expiry = get_config('factor_token', $expiryconfig);
             $expirystring = format_time($expiry);
-            $mform->addElement('advcheckbox', 'factor_token_trust', '', get_string('form:trust', 'factor_token', $expirystring));
-            $mform->setType('factor_token_trust', PARAM_BOOL);
-            $mform->setDefault('factor_token_trust', true);
+            if ($expiry !== 0) {
```
matthewhilton: the expiry check should probably move to line 151 IMO to avoid double nesting

e.g. 
```
$expiryconfig = is_siteadmin() ? 'adminexpiry' : 'expiry';
$expiry = get_config('factor_token', $expiryconfig);

if ($noproperty && $nostate && $weight !== 0 && !empty($expiry)) {
... add form elements
}
```

matthewhilton (changes requested):

PR #452, opened 2023-12-20
Closes #450. Fixed cohort factor list.

### LINKED ISSUE #450: Cohort factor lists only 25 cohorts.
Cohorts are retrieved here by a call of `cohort_get_all_cohorts()`
https://github.com/catalyst/moodle-tool_mfa/blob/MOODLE_400_STABLE/factor/cohort/settings.php#L41

The method retrieves by default only 25 cohorts, as defined here https://github.com/moodle/moodle/blob/6374475dc3cf3a94e22ec18c4655b89819a1b5d6/cohort/lib.php#L498 (cohort_get_all_cohorts($page = 0, $perpage = 25, $search = '', $withcustomfields = false)).

The settings page should include all the cohorts, not just the first 25.

haiksu:
Instead of using `cohort_get_all_cohorts()` method, which returns only 25 cohorts, this modification fetches all the cohorts straight from the database without limitations.

### DIFF
--- a/factor/cohort/settings.php
+++ b/factor/cohort/settings.php
@@ -38,10 +38,11 @@ $settings->add(new admin_setting_configtext('factor_cohort/weight',
     new lang_string('settings:weight', 'tool_mfa'),
     new lang_string('settings:weight_help', 'tool_mfa'), 100, PARAM_INT));
 
-$cohorts = cohort_get_all_cohorts();
+global $DB;
+$cohorts = $DB->get_records('cohort', null, 'name asc', 'id,name');
 $choices = [];
 
-foreach ($cohorts['cohorts'] as $cohort) {
+foreach ($cohorts as $cohort) {
     $choices[$cohort->id] = $cohort->name;
 }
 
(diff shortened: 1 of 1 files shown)


factor/cohort/settings.php
```diff
@@ -38,10 +38,11 @@
     new lang_string('settings:weight', 'tool_mfa'),
     new lang_string('settings:weight_help', 'tool_mfa'), 100, PARAM_INT));
 
-$cohorts = cohort_get_all_cohorts();
+global $DB;
```
Peterburnett: the global call isnt required here, as this file is direct included from a context where it should already be declared. This is only required inside of methods or other scope changes from the global scope.

haiksu: https://github.com/catalyst/moodle-tool_mfa/pull/452/commits/764f1e4e7c5ccc10dae1832f5620858003cf3ad2

danmarsden:
Thanks @haiksu - please note that as this plugin is now part of Moodle core - we need to address any issues/bugs/improvements in the upstream "latest" release before we backport to this repository to support older Moodle sites.
I note the current Moodle upstream release has the same "older" code:
https://github.com/moodle/moodle/blob/main/admin/tool/mfa/factor/cohort/settings.php#L41

If you end up creating an upstream Moodle tracker for this and fixing it there - please come back and post the details here so we can then look at merging this into the repo here for older sites - otherwise we need to close this as "won't fix."

PR #454, opened 2024-02-15
[#453] Ignore unknown factors when calculating passing score in grace…

### LINKED ISSUE #453: Grace mode quirk with factors that return UNKNOWN status
Login banner before being accepted returns an UNKNOWN status.

The grace factor when checking if other factors pass before redirecting, uses the `get_cumulative_weight` function. This, however, would stop checking factors if it encountered an `UNKNOWN` factor. See https://github.com/catalyst/moodle-tool_mfa/blob/29963a6fa7c6538c8ae79d924a034c03d6ff7755/classes/manager.php#L827-L829

So if you had a passing factor below a login banner factor, grace mode would trigger because it would never see the passing factor. So factors that require setup e.g. TOTP would get redirected to by grace mode, instead of passing outright.

matthewhilton:
Closes #453 

Have tested this locally with the setup:
- Login banner (0 points)
- Role factor (non-admin gets 100 points)
- TOTP
- Grace mode

Logging in as:

- Student - login banner is shown and once accepted they are logged in
- Admin - login banner is shown, and once accepted they get prompted for TOTP (assuming grace mode has expired)

Seems to work as expected. Also will be adding a unit tests for this scenario.

To Do:

- [x] Ci passing
- [x] Unit tests

### DIFF
--- a/factor/grace/classes/factor.php
+++ b/factor/grace/classes/factor.php
@@ -135,7 +135,7 @@ class factor extends object_factor_base {
                             }
 
                             // We should never redirect if we have already passed.
-                            if ($redirectable && \tool_mfa\manager::get_cumulative_weight() >= 100) {
+                            if ($redirectable && \tool_mfa\manager::get_total_weight() >= 100) {
                                 $redirectable = false;
                             }
 
--- a/factor/grace/tests/factor_test.php
+++ b/factor/grace/tests/factor_test.php
@@ -23,6 +23,7 @@ namespace factor_grace\tests;
  * @author      Peter Burnett <peterburnett@catalyst-au.net>
  * @copyright   Catalyst IT
  * @license     http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ * @covers      \factor_grace
  */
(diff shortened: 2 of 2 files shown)


factor/grace/classes/factor.php
```diff
@@ -135,7 +135,9 @@ public function get_state($redirectable = true) {
                             }
 
                             // We should never redirect if we have already passed.
-                            if ($redirectable && \tool_mfa\manager::get_cumulative_weight() >= 100) {
+                            // Do not stop on unknown, to avoid factors e.g. login banner from
+                            // hiding passing factors lower down in the list.
+                            if ($redirectable && \tool_mfa\manager::get_cumulative_weight(false) >= 100) {
```
Peterburnett: There is a get_total_weight method in the manager, and I think it duplicates the behaviour of get_cumulative with the stop on unknown flag = false, so I think this one can be simple. remove the stoponunknown flag, and call get_total_weight instead.

matthewhilton: Yep good catch - done

Issue #457, opened 2024-03-07
4.3 upgrade broken by higher version than core


aspark21:
https://github.com/catalyst/moodle-tool_mfa/blob/MOODLE_400_STABLE/factor/token/version.php#L29

![Screenshot 2024-03-07 at 17 18 39](https://github.com/catalyst/moodle-tool_mfa/assets/4015496/9cbb84d2-0a45-4a0e-8ce8-733462eb8989)

aspark21:
there are no upgrade steps in plugin or core so this can safely be worked around by updating the version in db before upgrade

samwitzig:
Thanks for mentioning the issue here. To make to upgrade easier (no fiddling around in the DB), I think the version of the plugin should be bumped in core. I created an issue: https://tracker.moodle.org/browse/MDL-81303

aspark21:
Practically yes, but those version numbers in core are aligned to the release date of the core version of Moodle

so it might go against core policy.

it’s a bit late now to revert the version number in the plugin since the current version has been rolled out across the board on catalyst eu managed instances

andrewnicols:
The version of this plugin should not have been updated in this plugin.

This will need to be downgraded in this plugin and instructions provided for how to fix it for those who already have the wrong version.

brendanheywood:
+1 to what @andrewnicols said, we need to manually back this out

dmitriim:
Would it be easier to fix it in core as a small bug tracker? so we avoid a pain for fixing it manually for potentially many people. Then we will never touch versions in this plugin again.

andrewnicols:
@dmitriim: We don't have a mechanism for doing this wihtout manual intervention.

dmitriim:
@andrewnicols Can we bump core version so it's higher than in the plugin?

brendanheywood:
We have a few hundred sites on this higher version, and yes that is our problem and we caused it and we can also fix that manually, BUT it also looks like there are many more hundreds of sites in the wild using that version of the plugin who will also hit this when upgrading to 4.3. As annoying as it is I think the best solution is to bump this in core. I can't see a programmatic way to fix this in code to downgrade itself.

andrewnicols:
I'll raise this with the integration team an see if we can support it with a higher version in Moodle 4.3.

Peterburnett:
[...]

PR #458, opened 2024-04-12
Fix factor token version #457

bwalkerl:
Based on the examples mentioned in #457 

Doesn't feel ideal to do this in version.php, but appears to be the least disruptive option.

### DIFF
--- a/factor/token/version.php
+++ b/factor/token/version.php
@@ -26,9 +26,14 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin->version      = 2023101600;      // The current plugin version (Date: YYYYMMDDXX).
+$plugin->version      = 2023080100;      // The current plugin version (Date: YYYYMMDDXX).
 $plugin->requires     = 2022041908; // Support Moodle 4.0 and higher.
 $plugin->component    = 'factor_token';
 $plugin->release      = 2022011700;
 $plugin->maturity     = MATURITY_STABLE;
 $plugin->dependencies = ['tool_mfa' => 2019102400];
+
+// Fix version numbers that are higher than 4.3 core.
+if (!during_initial_install() && (int) get_config('factor_token', 'version') >= 2023100900) {
+    set_config('version', '2023080100', 'factor_token');
+}


factor/token/version.php
```diff
@@ -26,9 +26,14 @@
...
-$plugin->version      = 2023101600;      // The current plugin version (Date: YYYYMMDDXX).
+$plugin->version      = 2023080100;      // The current plugin version (Date: YYYYMMDDXX).
 $plugin->requires     = 2022041908; // Support Moodle 4.0 and higher.
 $plugin->component    = 'factor_token';
 $plugin->release      = 2022011700;
 $plugin->maturity     = MATURITY_STABLE;
 $plugin->dependencies = ['tool_mfa' => 2019102400];
+
+// Fix version numbers that are higher than 4.3 core.
+if (!during_initial_install() && get_config('factor_token', 'version') >= '2023100900') {
```
Peterburnett: This string/int comparison doesnt feel quite right. Can you verify the type and do an explicit cast if required. Having it hardcoded as a string feels wrong.

Peterburnett (changes requested):
Comment around string/int comparison

factor/token/version.php
bwalkerl: Thanks @Peterburnett 

I copied from string part from the [example](https://github.com/catalyst/moodle-local_maillog/blob/MOODLE_35_STABLE/version.php) as get_config('factor_token', 'version') is returning a string, but that's a good point. 

This is running on every page so we'd want to go with the most efficient comparison. What do you think would be most efficient - leaving it as a string or casting with (int) or intval()?

factor/token/version.php
```diff
@@ -26,9 +26,14 @@
...
+    set_config('version', $plugin->version, 'factor_token');
```
danmarsden: using $plugin->version seems wrong here - we should be setting it specifically to 2023080100 in this case - so that if a future upgrade bumps to 2023080100.01 or 2023080101 then it downgrades to the right version and runs the upgrade steps that we might have to implement in future.

bwalkerl: Thanks @danmarsden - must have had a brain fart thinking through the update logic. I'll get that updated.

factor/token/version.php
bwalkerl: After some testing I ended up going with (int) get_config('factor_token', 'version') >= 2023100900 as the difference was almost negligible compared to the other check and getting the config.

PR #475, opened 2025-09-21
Remove erroneous apostrophe from `README.md`.

RokeJulianLockhart:
It shouldn't be possessive.


danmarsden:
Thanks for the PR - keep in mind this github repo is now bascially deprecated as the plugin is now part of the core Moodle release for 4.3 and higher releases.

PR #476, opened 2025-09-25
Only modify `$SESSION` when the plugin is actually enabled

cameron1729:


Peterburnett:
I think this is possible that this might hook during after_config, during which it might be modified by something like ABConfig, but I think this is good to go. CI is failing, which I havent checked

cameron1729:
Thanks @Peterburnett 

The CI failures don't seem related, and I can see from other MRs that CI has been failing in the repo for a while. I tried unit tests locally with this patch and they pass just fine.

I'll merge this one, cheers.

PR #488, opened 2026-09-10
Fix T20 tests

matthewhilton:
**Issue**
1. Settings classes moved to autoloaded namespaces, old ones emitting debugging during tests

**Fix**
- Switch to new classes

**Testing**
- Re-run testsuite, passing
- Manually went through tool_mfa + every factors settings page in gui to confirm no error or debugging
- Double checked every change to confirm its good (e.g. multi checkbox is still multi checkbox, and not just plain checkbox)

### DIFF
--- a/classes/local/admin_setting_managemfa.php
+++ b/classes/local/admin_setting_managemfa.php
@@ -16,6 +16,8 @@
 
 namespace tool_mfa\local;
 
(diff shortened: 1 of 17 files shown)


rhell4:
Everything looks good to me, thanks Matt!

